PR #12, opened 2026-02-27
feat: integrate benchmarker module — RAG quality evaluation

pahomm:
- Add src/metatron/benchmarker/ module (schemas, services, db, api)
- Add 6 metrics: correctness, relevancy, faithfulness, context precision/recall, confidence
- Add embedding_proxy service (OpenAI-compatible proxy for Ollama)
- Add migration 005_benchmarker (4 tables)
- Add benchmarker endpoints under /api/v1/benchmarker/
- Add return_trace parameter to hybrid_search_and_answer
- Add benchmarker optional dependencies in pyproject.toml
- Add 7 test files + conftest_benchmarker.py
- Update Dockerfile to include benchmarker deps
- Update docs (API, ARCHITECTURE, CONNECTORS, GETTING_STARTED, OBSERVABILITY)
- Fix ConnectionRow ORM to match migration 001 schema
- Add BENCHMARKER_EMBEDDING_PROXY_URL to .env.example and Settings


godedok:
1. Ollama is now mandatory — regression for all users                                                                                                                
                                                                                                                                                                       
  docker-compose.yml — profiles: ["ollama"] was removed and depends_on: ollama was added to the metatron service. Now docker compose up won't start without Ollama.    
  Most of us use DeepSeek/OpenRouter and don't need local Ollama running.

  Fix: restore profiles: ["ollama"], remove depends_on: ollama from metatron.

  2. Hard import of benchmark-qed breaks the app without optional deps

  api/app.py — the benchmarker router is imported at module level. The import chain pulls in benchmark-qed, which is declared as an optional dependency in
  pyproject.toml. If it's not installed, the entire API server fails to start — not just benchmarker, but everything (/health, /api/v1/chat, MCP, etc.).

  Fix: wrap the import in try/except ImportError or use a lazy import.

  3. runner.py blocks the event loop

  _run_single() calls hybrid_search_and_answer() (a synchronous function, 1-5 sec per call) directly from an async context. When running 100 questions, the entire API
  server is frozen for minutes — no health checks, no chat, no responses to any requests.

  Fix: wrap in await asyncio.to_thread(hybrid_search_and_answer, ...).

  4. No workspace_id validation in POST /test-runs/

  SaveTestRunRequest does not include workspace_id. The endpoint checks that benchmark_set_id exists but does not verify it belongs to the current workspace. A user
  from workspace "A" can write results to a benchmark set owned by workspace "B".

  Should fix

  5. logging instead of structlog — all 16 benchmarker files

  The rest of the project uses structlog (structured JSON logs, filterable by key). The benchmarker module uses stdlib logging (plain text). In production, benchmarker
   logs will be invisible to monitoring that expects JSON output.

  Mechanical replacement: import structlog / logger = structlog.get_logger().

  6. nest_asyncio.apply() in qed.py

  The documentation (BENCHMARKER_INTEGRATION.md) explicitly states: "nest_asyncio is not needed with this approach". But qed.py uses it and patches the global event
  loop. Either remove it from the code or update the documentation with a justification.

  7. embedding_proxy/Dockerfile — Python 3.11

  FROM python:3.11-slim — the project requires Python 3.12+. Should be python:3.12-slim.

  8. No tests for key modules

  Not covered by tests: runner.py (core orchestration of the entire benchmarking process), context_fetcher.py, POST /run-tests and POST /generate endpoints.

PR #13, opened 2026-03-05
feat: add dashboard API endpoints for analytics

pahomm:
- Add 5 dashboard endpoints: overview, sync-history, ingestion-errors, query-trend, graph-stats
- Add migration 006: sync_logs enhancements (source_title, qdrant_chunks, nullable connection_id)
- Add 28 unit tests with full coverage and graceful degradation
- Update API documentation with all endpoint specs
- Update architecture docs to mention dashboard routes
- Mark completed observability tasks in TODO

Endpoints:
- GET /api/v1/dashboard/overview - KPI cards (documents, jira issues, connectors)
- GET /api/v1/dashboard/sync-history - recent sync events with status
- GET /api/v1/dashboard/ingestion-errors - failed/partial syncs with error details
- GET /api/v1/dashboard/query-trend - daily query volume over time
- GET /api/v1/dashboard/graph-stats - knowledge graph statistics and lineage

All endpoints support graceful degradation and workspace isolation.


godedok:
Critical                                                                                                                                                             
                                                                                                                                                                       
  1. dashboard.py is 610 lines — 3x over the 200-line project limit                                                                                                    
                                                                                                                                                                       
  Should be split into at least 3 modules (e.g., dashboard/overview.py, dashboard/sync.py, dashboard/analytics.py), or extract the DB helper functions                 
  (_count_active_connectors, _get_sync_history, _get_ingestion_errors, _get_query_trend, _get_graph_stats) into a separate service/repository layer like
  storage/dashboard_queries.py.

  2. QueryTraceRow model has DateTime without timezone=True — mismatches migration 003

  Migration 003 creates query_traces.created_at as DateTime(timezone=True) with server_default=sa.func.now(), but the ORM model defines:
  created_at = Column(DateTime, default=datetime.utcnow, nullable=False)
  Missing timezone=True will cause timezone-aware vs. naive datetime issues at runtime. Same applies to SyncLogRow.

  3. Cypher f-string in _get_graph_stats — injection risk pattern

  f"... LIMIT {MAX_ORPHAN_NODES_LIMIT}"
  Currently safe (constant = 100), but f-strings in Cypher queries are a dangerous pattern. Should use parameterized query: LIMIT $limit with {"wid": workspace_id,
  "limit": MAX_ORPHAN_NODES_LIMIT}.

  4. store_query_trace in postgres.py — missing json.dumps(trace) for raw text() query

  The method uses text() with asyncpg. For JSONB columns in raw SQL, you should explicitly serialize with json.dumps(trace), same as store_document_version does for
  changed_fields. The ORM version in pg_connection.py handles this automatically, but the raw SQL version does not.

  Should fix

  5. _run_sync in connections.py has double commit

  Explicit session.commit() inside the function + get_session() context manager auto-commits on exit. Remove the explicit session.commit().

  6. documents_new, documents_updated, documents_skipped hardcoded to 0 in _run_sync

  The result object from ingest_documents already has result.documents_new, result.documents_skipped, etc. — these should be stored, not hardcoded as 0. Otherwise
  sync_logs will always show incomplete data.

  7. datetime.utcnow() is deprecated since Python 3.12

  Used in connections.py and ORM model defaults. Should be datetime.now(UTC). The store_query_trace_sync in pg_connection.py already uses the correct pattern — be
  consistent.

  8. Pydantic response models use str for datetime fields

  OverviewKPIResponse.last_upload: str | None and SyncHistoryItem.started: str should be datetime types — Pydantic v2 handles ISO 8601 serialization automatically.

  9. Dashboard endpoints access DB directly from the API layer

  dashboard.py directly imports get_session(), ORM models, and constructs raw Memgraph/Qdrant queries. This bypasses the storage abstraction layer (postgres.py,
  qdrant.py, memgraph.py). Dashboard should query through storage layer functions, not access databases directly from routes.

  10. Workspace validation duplicated 5 times

  Each endpoint repeats:
  manager = get_workspace_manager()
  workspace = manager.get_workspace(workspace_id)
  if workspace is None:
      raise HTTPException(status_code=404, ...)
  Should be extracted to a FastAPI dependency: Depends(get_valid_workspace).

  11. Missing tests for key new functionality

  No tests for: _run_sync sync log creation in connections.py, store_query_trace_sync in pg_connection.py, search.py trace logging additions, inner DB helper
  functions, error truncation logic.

  Minor

  12. Migration 006 downgrade will fail if NULL values exist — the comment acknowledges this but doesn't handle it.

  13. Unused import json in pg_connection.py — dead import.

  14. time.time() for timing in _run_sync — should use time.perf_counter() (monotonic, higher resolution).

  15. SyncHistoryItem.status uses # type: ignore[arg-type] — if sync log has unexpected status value (not "success"/"partial"/"failed"), Pydantic will raise a
  validation error at runtime.

  16. _get_graph_stats Cypher query without direction — (a)-[r]-(b) scans both directions, will be slow on large graphs. Use (a)-[r]->(b).

  17. Query trace logging is inline in the search hot path — store_query_trace_sync adds DB I/O to every search. Consider background task or fire-and-forget.

PR #26, opened 2026-03-13
Feature/clickable source links

pahomm:


godedok:
## Code Review: Clickable Source Links

Nice feature — URL propagation through the search pipeline is a solid improvement for UX. A few issues to address before merging.

---

### 🔴 Critical (blocking merge)

**1. Blocking I/O in async path**
`FileStore.save()` calls `file_path.write_bytes()` — synchronous disk I/O inside an `async def` endpoint. This blocks the entire event loop while writing large files. Same issue with `iterdir()` in the new download endpoint (`files.py:88-90`) — scans the whole workspace directory on every request.

Fix: wrap both in `asyncio.to_thread()`:
```python
await asyncio.to_thread(file_path.write_bytes, content)
```
```python
matches = await asyncio.to_thread(
    lambda: [f for f in ws_dir.iterdir() if f.name.startswith(f"{file_id}_")]
)
```

**2. Confluence `/wiki/` prefix breaks self-hosted instances**
`confluence.py:182` — hardcoding `/wiki/` is correct for Atlassian Cloud, but self-hosted Confluence Server/Data Center runs at the root. If `CONFLUENCE_URL` already includes `/wiki`, you get a double prefix.

Fix:
```python
wiki_prefix = "" if base_url.rstrip("/").endswith("/wiki") else "/wiki"
url = f"{base_url}{wiki_prefix}/spaces/{space_key}/pages/{page_id}"
```

---

### 🟡 Important (fix in this PR or immediate follow-up)

**3. Content-Disposition header injection**
`files.py:95-104` — `original_name` comes from user-supplied `file.filename` and goes into `FileResponse(filename=...)`. Should sanitize path separators and control characters:
```python
original_name = re.sub(r'[/\\<>:"|?*\x00-\x1f]', '_', original_name)
```

**4. `workspace_id=None` baked into Qdrant URLs**
`chat.py:339` — if `workspace_id` is None, you'll store `?workspace_id=None` in metadata. Guard:
```python
if file_id and workspace_id:
    url = f"/api/v1/files/{file_id}/download?workspace_id={workspace_id}"
else:
    url = ""
```

**5. Fragile test setup**
`test_file_download.py:28-36` — monkey-patching `config_module._settings` directly. Use `unittest.mock.patch` instead for thread safety and cleaner teardown.

---

### 🟢 Minor / Nits

**6.** `chat.py:336` — good catch fixing `"type": "confluence"` → `"type": "upload"` for uploaded files. Note: previously uploaded docs in Qdrant will still show the Confluence icon until re-indexed.

**7.** `pipeline.py:216` — the metadata spread order is correct (`doc.url` wins), but it's fragile if someone reorders. Worth an inline comment: `# must be after spread to take precedence`

**8.** `jira.py:129` — `self._config.get('url')` called twice. Minor, but cleaner as:
```python
base = self._config.get("url", "").rstrip("/")
url = f"{base}/browse/{issue_key}" if base else ""
```

**9.** `docs/superpowers/plans/2026-03-12-clickable-source-links.md` — 717-line plan doc will go stale quickly. Consider removing after the feature ships.

---

### Verdict: Request Changes

Two critical items need to be addressed — blocking I/O in async endpoints and the Confluence URL prefix. The rest can be a fast follow-up if needed.

PR #28, opened 2026-03-18
Feature/document level rbac

godedok:


pahomm:
Critical — blocks merge

  C1. Double-quoting bug in _acl_clause (graph_ops.py)                                                                                                                                                                                                             
  _esc() already wraps strings in single quotes ('value'), but _acl_clause adds another layer:                                                                                                                                                                  
  groups_list = "[" + ", ".join(f"'{_esc(g)}'" for g in user_groups) + "]"
  For "engineering" this produces [''engineering''] — invalid Cypher, query will crash at runtime. Fix: use _esc_list(user_groups) or drop the outer quotes.
                                                                                                                                                                                                                                                                
  C2. ACL bypass via search_by_date and search_by_type (search.py)                                                                                                                                                                                              
                                                                                                                                                                                                                                                                  search_with_date_filter has multiple code paths that call store.search_by_date() and store.search_by_type() without passing access_filter. When a date range or Jira query is detected, results come back completely unfiltered — documents leak to             unauthorized users.                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                
  C3. ACL bypass via entity_filter fallback (search.py)

  When combined_filter returns too few results, the code falls back to a fully unfiltered search:                                                                                                                                                                 unfiltered = store.hybrid_search(query, limit=k)
  return _merge_unique(filtered, unfiltered)[:k]                                                                                                                                                                                                                
  This completely defeats access_filter. When ACL is active, the fallback should relax title_filter only, never drop the access filter.
                                                                                                                                       
  C4. user_groups accepted but never used in get_graph_overview / get_graph_expand (graph_ops.py)                                                                                                                                                               
                                                                                                                                                                                                                                                                  Both functions accept user_groups in their signatures but the bodies never reference it — _acl_clause is never called. Graph endpoints will expose entities and relationships from documents the user should not have access to.                                                                                                                                                                                                                                                                                                C5. bcrypt missing from pyproject.toml                                                                                                                                                                                                                           
  passwords.py imports bcrypt but the package was never added to project dependencies. Fresh install will fail with ModuleNotFoundError.                                                                                                                           
  C6. Zero tests                                                                                                                                                                                                                                                
  
  +1886 lines of security-critical code (authentication, authorization, ACL filtering) with no tests at all. At minimum need unit tests for passwords.py, _acl_clause, _compose_filters, CRUD tests for UserStore, and integration tests for login and user       management endpoints.
                                                                                                                                                                                                                                                                  Important — should fix before merge                                                                                                                                                                                                                              
  I1. Race condition in seed_admin (user_store.py) — docstring says "Uses ON CONFLICT" but the implementation is SELECT COUNT then INSERT with no ON CONFLICT clause. Multiple replicas can create duplicate admins.                                               
  I2. asyncio.run() in _run_hooks_sync is fragile (search.py) — works only from thread pool threads. If hybrid_search_and_answer() is called from an async context directly, it will crash. No runtime guard exists.                                               
  I3. _resolve_user_groups only calls hooks[0] (graph.py) — graph endpoint calls only the first hook, while _run_hooks_sync in search chains all hooks. Inconsistent ACL behavior when multiple plugins are loaded.                                                
  I4. /api/v1/config not in PUBLIC_PATHS (middleware.py) — documented as "public, no auth required" but not added to PUBLIC_PATHS. With AUTH_ENABLED=true it will be behind the auth gate.                                                                         
  I5. Schema mismatch (user_store.py) — ensure_schema CREATE TABLE has no username column, but create_user INSERT references it. On a fresh DB before Alembic runs, the INSERT will fail. Also, raw DDL alongside Alembic creates a parallel schema management    path.
                                                                                                                                                                                                                                                                  I6. Missing guard for user_store on app.state (users.py) — UserStore init is wrapped in try/except, so if DB is unreachable app.state.user_store won't exist. Auth routes check with getattr, but user routes access it directly — will raise AttributeError.

PR #47, opened 2026-03-27
feat: user mapping Metatron <-> Open WebUI

pahomm:
- Personal API keys (api_keys table, mtk_ prefix, SHA-256 hashed)
- Per-user auth in /v1/chat/completions (replaces static key)
- API key management endpoints (POST/GET/DELETE /api/v1/users/{id}/api-keys)
- OpenWebUIClient for CRUD sync via Open WebUI REST API
- Bundled sync: auto-create users in Open WebUI on Metatron CRUD
- External import: POST /api/v1/admin/import-openwebui-users
- Conversation history from req.messages instead of in-memory store
- Config: METATRON_OPENWEBUI_URL, METATRON_OPENWEBUI_METATRON_URL

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -143,7 +143,9 @@ Frontend splits on `" — "` to extract URL; no URL → title only.
 - QUERY_EXPANSION_ENABLED (true) — LLM query expansion
 - GRAPH_EXTRACTION_ENABLED (true) — NER → Memgraph
 - METATRON_OPENAI_COMPAT_ENABLED (true) — OpenAI-compatible API for Open WebUI
-- METATRON_OPENAI_COMPAT_KEY ("") — static API key for OpenAI-compat endpoints (empty = disabled)
+- METATRON_OPENAI_COMPAT_KEY ("") — static API key for OpenAI-compat endpoints (Home scenario)
+- METATRON_OPENWEBUI_URL ("") — Open WebUI URL for bundled user sync
+- METATRON_OPENWEBUI_METATRON_URL ("") — external Metatron URL written into Direct Connections
 - See core/config.py for full list
 
 ## Open WebUI Integration
@@ -154,9 +156,16 @@ Endpoints:
 - `POST /v1/chat/completions` — chat completions (streaming + non-streaming)
 - `GET /v1/openapi.json` — stub for connection verification
 
-Auth: `Authorization: Bearer <METATRON_OPENAI_COMPAT_KEY>`
+Auth: personal API key (`mtk_...`) per user, or static `METATRON_OPENAI_COMPAT_KEY` fallback (Home scenario).
 
-Setup in Open WebUI: Settings → Connections → OpenAI API → Add URL `http://metatron:8000/v1` + key.
+Three deployment scenarios:
+1. **Home** — single user, no auth, static API key via global Open WebUI connection
+2. **Bundled** — multi-user, Metatron syncs users to Open WebUI, each gets personal API key + Direct Connection
+3. **External** — import users from existing Open WebUI via `POST /api/v1/admin/import-openwebui-users`
+
+In Bundled/External, `ENABLE_DIRECT_CONNECTIONS=true` is mandatory in Open WebUI. Without it users share a global API key and can spoof each other's identity.
+
+Bundled sync: on startup Metatron auto-registers `admin@metatron.local` in Open WebUI. On user CRUD, changes are mirrored. OWUI admin password is the same as `AUTH_PASSWORD` (default: `metatron`).
 
 Docker: Open WebUI available in `docker-compose.full.yml` with profile `openwebui` on port 3080.
 
--- a/docs/API.md
+++ b/docs/API.md
@@ -1336,7 +1336,7 @@ Delete a test run and all its results.
 
 OpenAI-compatible endpoints for connecting Open WebUI or any OpenAI-compatible client to Metatron.
 
-**Authentication:** `Authorization: Bearer <METATRON_OPENAI_COMPAT_KEY>`
+**Authentication:** `Authorization: Bearer <personal API key (mtk_...)>` or `Bearer <METATRON_OPENAI_COMPAT_KEY>` (Home scenario fallback)
 
 **Error format:** `{"error": {"message": "...", "type": "invalid_request_error"}}`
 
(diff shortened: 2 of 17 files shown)


godedok:
Code Review: PR #47 — feat: user mapping Metatron <-> Open WebUI                                                                                          
                                                                                                                                                            
  PR: aisec-co-il/metatroncore#47                                                                                                                           
  Branch: feature/user-mapping-openwebui → develop                                                                                                          
  Scope: +1116 / -50 across 15 files (6 new modules, 5 new test files, 4 modified)                                                                          
  Tests: 31 passed, 7 need respx dep, 0 failed                                                                                                              
                                                                                                                                                            
  Summary                                                                                                                                                   
                                                                                                                                                            
  Replaces the static METATRON_OPENAI_COMPAT_KEY auth with personal API keys (mtk_ prefix, SHA-256 hashed in DB). Adds bidirectional user sync between      
  Metatron and Open WebUI: bundled sync (auto-create users in OWUI on Metatron CRUD) and external import (POST /api/v1/admin/import-openwebui-users).
  Conversation history now extracted from req.messages instead of an in-memory store.                                                                       
                  
  Strengths

  1. SHA-256 hashed keys, raw key shown once — correct pattern. DB leak does not expose raw keys. key_prefix (first 12 chars) stored for identification.    
  Clean API: create_key → returns raw, resolve_key → hash lookup.
  2. Backward compatible auth — verify_openai_compat_key tries personal key first, falls back to static key, falls back to legacy hmac.compare_digest.      
  Zero-downtime migration path.                                                                                                                             
  3. In-memory history store removed — _conversation_history, _history_lock, _record_history() all deleted. History now extracted from req.messages (sent by
   Open WebUI). Correct — Open WebUI already maintains conversation state; Metatron was duplicating it.                                                     
  4. Clean separation of concerns — api_key_store.py (storage), openwebui_client.py (HTTP client), openwebui_sync.py (orchestration), openwebui_import.py
  (admin endpoint). Each module has a single responsibility.                                                                                                
  5. Graceful degradation everywhere — OWUI sync failures are logged and swallowed (user creation in Metatron still succeeds). api_key_store unavailable →
  fallback to static key. openwebui_url empty → sync disabled.                                                                                              
  6. ensure_admin signup/signin flow — handles both fresh OWUI (signup → first user = admin) and existing OWUI (signin). Idempotent.
  7. Good test coverage — API key CRUD, import with skips, sync lifecycle, ensure_admin flows. SQLite in-memory for fast isolated tests.                    
                                                                                                                                                            
  Issues                                                                                                                                                    
                                                                                                                                                            
  Critical        

  1. owui_token (Open WebUI JWT) stored in plaintext in users table (user_store.py:872-873) — owui_token is a valid authentication token for the Open WebUI 
  instance. Storing it unencrypted means a Postgres breach exposes full OWUI access for every synced user. This should be Fernet-encrypted (same pattern as
  connector credentials) or not stored at all (re-authenticate when needed via _ensure_login).                                                              
  2. Import endpoint returns raw passwords and API keys in response (openwebui_import.py:305-311) — metatron_password and api_key are returned in the JSON
  response body. If this response is logged (middleware, reverse proxy access logs, audit logs), credentials leak. Consider: (a) return passwords only once,
   clearly marked, or (b) force a password reset flow instead of generating passwords server-side.
  3. Import endpoint has no RBAC via standard middleware (openwebui_import.py:251) — uses getattr(request.state, "user", {}) manual check instead of        
  Depends(require_admin). If AUTH_ENABLED=False, request.state.user won't exist, and {}.get("role") returns None → 403 (safe). But this bypasses the        
  standard auth dependency chain. Use Depends(require_admin) for consistency and to prevent future regression.
                                                                                                                                                            
  Important       

  4. revoke_key by key_prefix can delete wrong key (api_key_store.py:571-579) — key_prefix is the first 12 chars of the raw key (mtk_ + 8 hex chars). With  
  32-bit entropy in the prefix, collision probability is low but nonzero. More importantly, if a user has multiple keys that happen to share a prefix
  (extremely unlikely with secrets.token_hex but architecturally wrong), DELETE WHERE key_prefix = :prefix could delete multiple keys. Use the id column    
  instead:        
  DELETE FROM api_keys WHERE id = :key_id AND user_id = :user_id
  5. resolve_key does static key comparison before hash (api_key_store.py:537-538) — if static_key and raw_key == static_key is a non-constant-time         
  comparison. The original code used hmac.compare_digest. While timing attacks on local comparisons require very specific conditions, this is a regression  
  from the security baseline. Use hmac.compare_digest for the static key check.                                                                             
  6. OpenWebUIClient httpx client never closed in lifespan (app.py:44-48) — owui_sync stores an OpenWebUIClient with an httpx.AsyncClient, but there's no   
  cleanup in the lifespan shutdown path. Add await owui_sync._client.close() to the shutdown section, or use @asynccontextmanager pattern.                  
  7. _user_engine created outside try/except (app.py:8-10) — create_async_engine(settings.postgres_dsn) is moved before the user_store try block. If
  postgres_dsn is malformed, this crashes the entire lifespan with an unhandled exception. Wrap it in its own try/except.                                   
  8. No migration for api_keys table — ensure_schema() uses CREATE TABLE IF NOT EXISTS, which works for new installs but doesn't use Alembic. This is
  consistent with user_store (which uses the same pattern), but both should eventually be migrated to proper Alembic migrations.                            
                  
  Minor                                                                                                                                                     
                  
  9. respx not in dependencies — test_openwebui_client.py imports respx but it's not in pyproject.toml. Tests fail with ModuleNotFoundError: No module named
   'respx'. Add respx to test dependencies.
  10. admin_password=settings.auth_password (app.py:42) — this couples the OWUI admin password to METATRON_AUTH_PASSWORD. If someone changes the Metatron   
  auth password, the OWUI sync breaks. Consider a separate METATRON_OPENWEBUI_ADMIN_PASSWORD config, or document this coupling.                             
  11. _map_role_from_owui maps everything non-admin to viewer (openwebui_import.py:239) — OWUI's "user" role becomes Metatron's "viewer", not "editor". This
   may be intentional (conservative default), but it means imported OWUI users can't edit documents without manual role promotion.                          
  12. No rate limiting on resolve_key — every /v1/chat/completions call does a DB query to resolve the API key. At scale, this could be a bottleneck.
  Consider an in-memory TTL cache (e.g., cachetools.TTLCache) for resolved keys.                                                                            
  13. get_user_by_id now returns owui_user_id and owui_token (user_store.py:891-893) — these fields are included in the SELECT and will appear in API
  responses from GET /api/v1/users/{id}. The OWUI token should not be exposed in API responses. Filter it out in the route handler or exclude it from the   
  SELECT.         
                                                                                                                                                            
  Architecture Assessment

  The PR sits correctly in the architecture:                                                                                                                
  - api_key_store.py and openwebui_client.py are auth layer (L3)
  - openwebui_sync.py orchestrates between auth and external system (L3)                                                                                    
  - Routes in api/routes/ (L6) use DI from app.state                    
  - No upward dependency violations                                                                                                                         
                                                                                                                                                            
  The user mapping resolves the user_mapping.py: NotImplementedError TODO that's been in the codebase.                                                      
                                                                                                                                                            
  The conversation history change is a significant improvement — the old in-memory store was unbounded (despite _MAX_HISTORY_USERS), not horizontally       
  scalable, and duplicated what Open WebUI already manages.                                                                                                 
                                                                                                                                                            
  Verdict         

  Request changes. The plaintext owui_token storage (issue #1) is a security issue that should be fixed before merge. The raw credentials in import response
   (issue #2) and non-standard RBAC check (issue #3) are also blocking. The respx missing dependency (issue #9) breaks CI. Everything else can be follow-up

PR #72, opened 2026-04-10
feat(WS1): add Redis session cache + MemoryService skeleton (Stage 2)

pahomm:
- storage/memory_redis.py: RedisSessionCache with TTL-bound session CRUD
- agent/memory_service.py: MemoryService skeleton with Redis + Neo4j write-through
- promote()/save() stub as NotImplementedError (needs PG + Qdrant)
- 25 unit tests, all passing

### DIFF
--- a/src/metatron/agent/memory_service.py
+++ b/src/metatron/agent/memory_service.py
@@ -1,0 +1,144 @@
+"""MemoryService — orchestrates agent memory across stores (WS1).
+
+Sits at L4 (agent layer), composes L1 storage modules:
+  - RedisSessionCache  (session memory — hot cache with TTL)
+  - memory_graph.py    (Neo4j — relationships and graph queries)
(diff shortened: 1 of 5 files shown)


godedok (approved):
LGTM — clean code, proper layering, 49/49 tests pass.

**Non-blocking notes for follow-up stages:**

1. `RedisSessionCache` doesn't formally inherit `SessionMemoryInterface` — makes sense since `promote()` lives in `MemoryService`. Consider having `MemoryService` implement the ABC once `promote()` is fully wired.

2. `cache()` index update (GET → append → SET) is not atomic — documented, fine for single-agent sessions. Migrate to Redis SADD when scaling.

3. `save_memory_to_graph()` is called without `entity_names`/`document_ids` for session memory — intentional, but worth an inline comment explaining why.

PR #75, opened 2026-04-14
feat(MTRNIX-247): memory hybrid search service (WS1 Stage 3)

godedok:
## Summary

WS1 Stage 3: implements `MemorySearchService` — a hybrid memory search that fans
out across Qdrant (vector), Neo4j (graph), and Redis (session) legs in parallel,
then blends results with configurable weights.

- Dense: min-max-normalized Qdrant RRF score (× `dense` weight, default 0.6)
- Graph: Neo4j `importance_score` for agent-pinned memories (× `graph` weight, default 0.3)
- Session: boolean boost for records found in the session Redis cache (× `session` weight, default 0.1)

Legs degrade gracefully via `retrieval.fallback._safe_call` — a down Neo4j or Redis
does not fail the search. Graph-only hits with no content are dropped.

## Files touched

**New:**
- `src/metatron/memory/__init__.py`
- `src/metatron/memory/search.py` — `MemorySearchService`, `MemorySearchWeights`
- `tests/unit/test_memory_search.py` — 12 tests

**Modified:**
- `src/metatron/core/config.py` — 4 new env vars: `METATRON_MEMORY_SEARCH_DENSE_WEIGHT`,
  `METATRON_MEMORY_SEARCH_GRAPH_WEIGHT`, `METATRON_MEMORY_SEARCH_SESSION_WEIGHT`,
  `METATRON_MEMORY_SEARCH_TOP_K_MULTIPLIER`
- `src/metatron/agent/memory_service.py` — optional `search=` kwarg + `MemoryService.search()` delegate
- `tests/unit/test_memory_service.py` — 2 new delegation tests

## DoD

- [x] `MemorySearchService.hybrid_search()` returns ranked `list[MemorySearchResult]`
- [x] Three legs fan out via `asyncio.gather(return_exceptions=False)` with per-leg `_safe_call`
- [x] Dedup by `record_id` across legs; per-leg sub-scores retained
- [x] Tags post-filter (in-process; no Qdrant tags index this iteration)
- [x] Scope passed as `.value` string to Qdrant + Neo4j
- [x] `MemoryService.search()` delegates to injected `MemorySearchService`; raises when missing
- [x] Backward compatible — existing `MemoryService` tests untouched
- [x] Ruff lint + format clean on touched files
- [x] mypy: no new errors introduced on touched files
- [x] All 28 unit tests (12 new search + 16 service) pass

## Test plan

- [ ] `pytest tests/unit/test_memory_search.py tests/unit/test_memory_service.py -v`
- [ ] Confirm no regressions in full unit suite outside pre-existing failures

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 - feat: WS1 Stage 1 — core memory models and interfaces (MTRNIX-240)
+- feat: memory hybrid search service (`MemorySearchService`) combining Qdrant vector, Neo4j graph, and Redis session legs with weighted blend and graceful degradation (MTRNIX-247)
 - **Breaking**: Replace Memgraph with Neo4j Community Edition for knowledge graph storage
   - Docker image: `memgraph/memgraph:2.18.1` → `neo4j:5-community`
   - Env vars: `MEMGRAPH_*` → `NEO4J_*` (old names still work via aliases)
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -87,6 +87,7 @@ src/metatron/
 ├── observability/             # health.py, metrics.py, tracer.py
 ├── workspaces/                # L3 — manager.py, models.py, persistence.py
 ├── skills/                    # L3 — engine.py
+├── memory/                    # L3 — hybrid search across Qdrant + Neo4j + Redis (MemorySearchService)
 ├── benchmarker/               # L2 — api/, db/, schemas/, services/metrics/
 └── scripts/                   # graph_audit.py, run_eval.py, grid_search_weights.py,
                                # graph_rebuild.py, graph_process.py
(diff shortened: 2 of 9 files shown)


pahomm:
---
  WARNING (should fix before merge)

  [WARNING] src/metatron/memory/search.py:17 — Importing private _safe_call from another module

  from metatron.retrieval.fallback import _safe_call

  Dependency direction is valid (L3 → L2), but _safe_call is a private symbol (_ prefix) of the retrieval module. Other modules should not depend on internal APIs that can change without notice. Additionally, the lack of type annotations on _safe_call     
  forces 3 # type: ignore[no-untyped-call] suppressions (lines 79, 89, 102).

  Recommendation: Either (a) promote safe_call to a public utility in core/utils.py / core/resilience.py with proper types, or (b) create a local 10-line equivalent in memory/search.py — the function is trivial.

  ---
  [WARNING] src/metatron/memory/search.py:237 — _hydrate_graph_record rebuilds lookup dict on every call

  def _hydrate_graph_record(node, workspace_id, session_hits):
      session_lookup = {r.id: r for r in session_hits}  # O(m) every time

  This function is called in a loop over graph_hits (line 149). Total complexity: O(n × m) instead of O(n + m).

  Fix: Build session_lookup once in hybrid_search() and pass it instead of session_hits:

  session_lookup = {r.id: r for r in session_hits}
  # ...
  hydrated = _hydrate_graph_record(node, workspace_id, session_lookup)

  ---
  SUGGESTION (can be addressed later)

  [SUGGESTION] search.py:262 — _min_max_normalize(all_ids: Any) typed as Any instead of Iterable[str]. Violates mypy strict.

  [SUGGESTION] search.py:111 — return_exceptions=False is the default for asyncio.gather, explicit argument is redundant.

  [SUGGESTION] models.py:337 — sparse_score in MemorySearchResult is never populated (always 0.0). Qdrant RRF fusion doesn't return a separate sparse score. Either document as reserved or remove.

  [SUGGESTION] search.py:27-32 — MemorySearchWeights doesn't validate that weights sum to 1.0. Custom env vars can produce scores > 1.0. Consider adding __post_init__ with a warning.

  [SUGGESTION] Tests — Missing coverage for: (a) Qdrant leg failure with graph+redis still returning results, (b) all three legs failing simultaneously, (c) three-way dedup (same record_id from all three sources).

PR #76, opened 2026-04-14
feat(MTRNIX-248): memory PostgreSQL store + MemoryService orchestrati…

pahomm:
…on (WS1 Stage 2-3)

Adds PostgreSQL as the source of truth for agent memory and wires the MemoryService orchestrator across Redis + Qdrant + PG + Neo4j.

- migration 013: memory_records + memory_snapshots tables with ws/agent/ scope/ttl indexes and GIN on tags
- storage/memory_postgres.py: MemoryPostgresStore with CRUD, dedup-by-hash and TTL cleanup; snapshot metadata CRUD
- agent/memory_service.py: compose PG + Qdrant + Redis + Neo4j with content-hash dedup, PG-first writes, best-effort Neo4j writes, Redis→PG session fallback, promote() that upgrades existing record scope on dedup hit, and reset() that enumerates IDs first to keep derived stores consistent across scope-only and agent-only filters
- CLAUDE.md updates in core/, storage/, agent/ reflecting current model shapes, scope values and interface method names

### DIFF
--- a/migrations/versions/013_memory_system.py
+++ b/migrations/versions/013_memory_system.py
@@ -1,0 +1,117 @@
+"""Add memory_records and memory_snapshots tables for WS1.
+
+Revision ID: 013
+Revises: 012
+Create Date: 2026-04-14
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+import sqlalchemy as sa
+from alembic import op
+
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+
+revision: str = "013"
+down_revision: str | None = "012"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "memory_records",
+        sa.Column("id", sa.Text, primary_key=True),
+        sa.Column("workspace_id", sa.Text, nullable=False),
+        sa.Column("agent_id", sa.Text, nullable=False),
+        sa.Column("scope", sa.Text, nullable=False),
+        sa.Column("source_type", sa.Text, nullable=False, server_default=""),
+        sa.Column("content", sa.Text, nullable=False, server_default=""),
+        sa.Column(
+            "tags",
+            sa.dialects.postgresql.JSONB,
+            nullable=False,
+            server_default=sa.text("'[]'::jsonb"),
+        ),
+        sa.Column(
+            "importance_score",
+            sa.Float,
+            nullable=False,
+            server_default=sa.text("0.5"),
+        ),
+        sa.Column("ttl_expires_at", sa.DateTime(timezone=True), nullable=True),
+        sa.Column("content_hash", sa.Text, nullable=False, server_default=""),
+        sa.Column("session_id", sa.Text, nullable=True),
+        sa.Column(
+            "metadata",
+            sa.dialects.postgresql.JSONB,
+            nullable=False,
+            server_default=sa.text("'{}'::jsonb"),
+        ),
+        sa.Column(
+            "created_at",
+            sa.DateTime(timezone=True),
+            nullable=False,
+            server_default=sa.text("NOW()"),
+        ),
+        sa.Column(
+            "updated_at",
+            sa.DateTime(timezone=True),
+            nullable=False,
+            server_default=sa.text("NOW()"),
+        ),
+    )
+
+    op.create_index("ix_memory_records_workspace", "memory_records", ["workspace_id"])
+    op.create_index("ix_memory_records_agent", "memory_records", ["workspace_id", "agent_id"])
+    op.create_index("ix_memory_records_scope", "memory_records", ["workspace_id", "scope"])
+    op.create_index(
+        "ix_memory_records_ttl",
+        "memory_records",
+        ["workspace_id", "ttl_expires_at"],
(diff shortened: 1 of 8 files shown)


godedok:
## Review summary

Tests pass (52/52 new), lint clean (except one formatting nit), mypy clean on touched files, layer boundaries respected. Approving with changes — several correctness and isolation concerns should be addressed before merge.

### Process / scope
- Commit message and PR title reference `MTRNIX-248`, but the scope (MemoryService orchestration + PG store) matches `MTRNIX-246` per its Jira description and the comment *"The scope of postgres will be done in this task"*. Please fix the key in the PR title and commit message.
- MTRNIX-246 DoD requires integration tests — only unit tests are present. Add integration coverage against real PG/Qdrant/Neo4j (services are assumed running per CLAUDE.md).

### Important
1. **Workspace-isolation gap in orchestration.** `MemoryQdrantStore` binds a single `workspace_id` at construction, but `MemoryService.save/delete/reset` take `workspace_id` as a parameter and pass records through without verifying `record.workspace_id == self._qdrant._workspace_id`. Reusing a `MemoryService` across workspaces would silently route writes into the wrong collection. Either drop the `workspace_id` parameter from service methods (it's implicit in the per-workspace stores) or assert equality on entry. Add a unit test that fails on mismatch. (`src/metatron/agent/memory_service.py:174–248`, `src/metatron/storage/memory_qdrant.py:50–63`)

2. **TTL index scope + missing trigger.** Partial index `ix_memory_records_ttl ON (ttl_expires_at) WHERE ttl_expires_at IS NOT NULL` is global, but `delete_expired(workspace_id)` filters by workspace. Make the index composite `(workspace_id, ttl_expires_at) WHERE ttl_expires_at IS NOT NULL`. Also, nothing currently calls `delete_expired` — document who triggers it (cron/event/periodic job) or it becomes dead code. (`migrations/versions/013_memory_system.py:74–79`, `src/metatron/storage/memory_postgres.py:240–253`)

3. **Input-record mutation + no content normalization.** `MemoryService.save/promote` overwrite `record.content_hash` on a caller-owned dataclass. Hash is computed on raw content, so `"hello"` and `"hello "` don't dedupe. Either normalize before hashing (strip/lowercase or `core.utils.normalize_text`) or document the exact-match semantics in the docstring and add a test. (`src/metatron/agent/memory_service.py:179, 263`)

4. **Race in `reset()`.** Listing IDs → `DELETE` → per-id Qdrant/Neo4j cleanup allows concurrent inserts between list and delete, leaving orphans in derived stores — exactly the bug the enumeration was meant to prevent. Fix with `DELETE ... RETURNING id` inside a transaction so the deleted-id set is authoritative. (`src/metatron/agent/memory_service.py:213–248`)

5. **Partial failure in `promote()` dedup branch.** If PG save succeeds and Qdrant upsert raises, the Redis source record is not cleaned up and the in-memory `existing.scope` has already been mutated. Add a test covering promote-with-dedup-and-qdrant-failure; either retry-safe cleanup or document the recovery contract. (`src/metatron/agent/memory_service.py:280–292`)

6. **Breaking constructor change.** `MemoryService.__init__` now requires `pg_store`. `api/dependencies.py` on `develop` constructs it without this argument — will raise `TypeError` at runtime as soon as anyone calls `get_memory_service()`. Update `dependencies.py` in this PR, or make `pg_store: MemoryPostgresStore | None = None` and raise only when PG-requiring methods are invoked. Call this out in CHANGELOG.

### Minor
7. `migrations/versions/013_memory_system.py` fails `ruff format --check`. Run `make format` and recommit.
8. No unique constraint or index on `(workspace_id, agent_id, content_hash)` — dedup returns "most recent" via `ORDER BY created_at DESC LIMIT 1`. Fine for small N, degrades as memory grows. Add an index or a unique constraint. (`src/metatron/storage/memory_postgres.py:215–229`)
9. `MemorySnapshot` table has no `delete_snapshot` method — snapshots accumulate forever. Either add deletion or document immutability and add a `(workspace_id, agent_id, label)` uniqueness rule.
10. `qdrant_store.upsert(existing)` in the dedup-promote branch re-runs embedding even though content is unchanged. Prefer `set_payload`-style scope update if available. (`src/metatron/agent/memory_service.py:285`)

### Nits
- The `builtins.list[...]` aliasing workaround (`memory_postgres.py:24`, `memory_service.py:24`) is ugly. Consider renaming the PG-store method to `list_records` / `list_snapshots` for parity with `list_session`.
- Test suite is unit-only (`MagicMock` engine + custom `_FakeCtx`). Add `tests/integration/test_memory_postgres_integration.py` to exercise real JSONB casts, GIN tags index, partial TTL index, `ON CONFLICT`, and `RETURNING` (if adopted).

### Well done
- Clean L1/L4 separation; no upward imports.
- Best-effort Neo4j writes via `asyncio.to_thread` + structlog warning.
- Thoughtful `reset()` design (despite the race in #4) — awareness of the over-delete bug in `delete_by_agent`.
- Solid failure-path coverage in tests (`pg_failure_propagates`, `qdrant_failure_propagates`, `neo4j_failure_does_not_block_save`).
- CLAUDE.md updates accurately reflect new method signatures.

### Blocking for merge
1. Fix PR title / commit message key: `MTRNIX-248` → `MTRNIX-246`.
2. Run `make format` (Minor #7).
3. Address Important #1, #4, #6 before merge.
4. Add integration tests (DoD requirement).

PR #77, opened 2026-04-14
feat(MTRNIX-248): memory REST API endpoints (WS1 Stage 3)

godedok:
## Summary

WS1 Stage 3: exposes 4 memory operations via REST under `/api/v1/memory/*`. Workspace is always derived from the authenticated user — never accepted from request body or query.

**Rebased on top of MTRNIX-246 (PR #76 / commit `fea154e`)**: `MemoryService` is now PG-backed with content-hash dedup. This PR composes its router + DI on the new orchestrator — no duplicate service methods.

## Endpoints

| Method | Path | RBAC | Service call |
|---|---|---|---|
| `POST` | `/records` | editor+ | `MemoryService.save` (PG dedup → PG → Qdrant → Neo4j) or `cache_session` (SESSION scope → Redis + Neo4j) |
| `POST` | `/search` | viewer+ | `MemoryService.search` → `MemorySearchService.hybrid_search` |
| `GET`  | `/records` | viewer+ | `MemoryService.list_records` (PG-backed) or `list_session` (Redis) |
| `DELETE` | `/records/{id}` | editor+ | `MemoryService.delete` (PG authoritative → Qdrant + Neo4j best-effort) |

## Files touched (after rebase)

**New:**
- `src/metatron/api/routes/memory.py` — router + 6 pydantic v2 schemas (312 lines)
- `src/metatron/memory/serde.py` — shared `record_from_qdrant_payload` helper
- `tests/unit/api/__init__.py`, `tests/unit/api/test_memory_routes.py` — 14 tests incl. `TestMemoryCRUDCycle` (full CRUD via TestClient)

**Modified:**
- `src/metatron/storage/memory_qdrant.py` — adds `get(record_id)` and `scroll(...)`; `delete()` returns `bool` (existence pre-check via `get`)
- `src/metatron/auth/dependencies.py` — adds `require_viewer` guard mirroring `require_editor`/`require_admin`
- `src/metatron/api/dependencies.py` — adds `get_memory_service` with per-workspace cache; builds shared `MemoryPostgresStore` + AsyncEngine on `app.state`
- `src/metatron/api/app.py` — registers memory router under `/api/v1`
- `src/metatron/memory/__init__.py`, `memory/search.py` — promote `_record_from_qdrant` → `serde.record_from_qdrant_payload`
- `tests/unit/test_memory_qdrant.py` — +1 test (delete returns False on miss)

## DoD

- [x] All 4 endpoints operational (`POST /records`, `POST /search`, `GET /records`, `DELETE /records/{id}`)
- [x] Pydantic v2 request/response schemas (auto-documented via FastAPI OpenAPI)
- [x] RBAC enforced (editor for write/delete, viewer for read)
- [x] `workspace_id` always derived from auth, never from request
- [x] Integration test for full CRUD cycle (`TestMemoryCRUDCycle::test_create_search_list_delete`)
- [x] Backward compatibility: no public signature changes to pre-existing `MemoryService` methods (rebased on MTRNIX-246's PG orchestrator)
- [x] Ruff + mypy clean on touched files
- [x] 63 memory tests pass; full unit suite +23 passed, 0 regressions vs baseline

## Test plan

- `pytest tests/unit/api/ tests/unit/test_memory_service.py tests/unit/test_memory_qdrant.py` → 63 passed
- `make lint && make typecheck` → clean on touched files
- Verification curls (manual, requires running services):
  - `curl -X POST /api/v1/memory/records -d '{...}'` returns 201
  - `curl -X POST /api/v1/memory/search -d '{"query":"..."}'` returns 200 with results

## Notes

- **Event emission deferred**: `MEMORY_STORED` / `MEMORY_DELETED` constants in `core/events.py` exist but no handlers subscribe. An earlier draft of this PR emitted them; dropped during rebase because MTRNIX-246 did not add emission and no consumer exists yet. Will be added as a follow-up once a subscriber (CC audit / activity timeline) lands.
- **DELETE semantics** (from MTRNIX-246): PG is source of truth — 404 when PG does not have the record; Qdrant failure logged not propagated; Neo4j best-effort. Session records are managed separately via `invalidate_session`.
- **Listing source**: PG `list_records` (not Qdrant scroll) — PG is the source of truth after MTRNIX-246.
- No changes to `core/interfaces.py`, `core/events.py`, `core/models.py`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - feat: WS1 Stage 1 — core memory models and interfaces (MTRNIX-240)
 - feat: memory hybrid search service (`MemorySearchService`) combining Qdrant vector, Neo4j graph, and Redis session legs with weighted blend and graceful degradation (MTRNIX-247)
+- feat: memory REST API endpoints (`POST /api/v1/memory/records`, `POST /api/v1/memory/search`, `GET /api/v1/memory/records`, `DELETE /api/v1/memory/records/{id}`) with workspace scope, RBAC, and full CRUD integration test (MTRNIX-248)
 - **Breaking**: Replace Memgraph with Neo4j Community Edition for knowledge graph storage
   - Docker image: `memgraph/memgraph:2.18.1` → `neo4j:5-community`
   - Env vars: `MEMGRAPH_*` → `NEO4J_*` (old names still work via aliases)
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -87,7 +87,8 @@ src/metatron/
 ├── observability/             # health.py, metrics.py, tracer.py
 ├── workspaces/                # L3 — manager.py, models.py, persistence.py
 ├── skills/                    # L3 — engine.py
-├── memory/                    # L3 — hybrid search across Qdrant + Neo4j + Redis (MemorySearchService)
+├── memory/                    # L3 — search.py (hybrid MemorySearchService), serde.py (shared Qdrant payload deserializer)
+│                              #   Orchestration lives in agent/memory_service.py (PG source of truth)
 ├── benchmarker/               # L2 — api/, db/, schemas/, services/metrics/
 └── scripts/                   # graph_audit.py, run_eval.py, grid_search_weights.py,
                                # graph_rebuild.py, graph_process.py
@@ -131,7 +132,7 @@ Request → OptionalAuthMiddleware (if AUTH_ENABLED)
 → auth/dependencies.py: get_current_user()
   → 1. Check plugin_manager.get_auth_provider()
   → 2. Fallback: jwt.verify_token()
-→ require_admin() / require_editor() check role level
+→ require_admin() / require_editor() / require_viewer() check role level
 ```
 
 ## Search Pipeline
--- a/src/metatron/api/.claude/claude.md
+++ b/src/metatron/api/.claude/claude.md
@@ -142,6 +142,22 @@ OpenAI-compatible API endpoints for Open WebUI integration:
 `GET /api/v1/finops/time-savings` — time savings metrics.
 See `routes/.claude/finops.md` for full documentation.
 
+### `routes/memory.py`
+Memory REST API endpoints (workspace-scoped, RBAC-gated):
+
+| Method | Path | RBAC | Purpose |
+|--------|------|------|---------|
+| POST | `/api/v1/memory/records` | editor+ | Create record — `service.save()` (PG→Qdrant→Neo4j) or `service.cache_session()` for SESSION scope. Returns 201 |
+| POST | `/api/v1/memory/search` | viewer+ | Hybrid search via `MemorySearchService.hybrid_search()`. 503 if search not configured |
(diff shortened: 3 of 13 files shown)


pahomm:
IMPORTANT

  1. Duplicated workspace resolution logic

  Two identical functions:
  - dependencies.py:48 — _resolve_workspace_id()
  - routes/memory.py:263 — _get_workspace_id()

  Same logic, same fallback chain. If one is updated, the other will be forgotten. Should be a single function imported from dependencies.py.

  2. assert in production code — routes/memory.py:332:
  assert body.session_id is not None
  python -O strips assertions. The model validator guarantees this, but type narrowing should use an explicit check:
  if body.session_id is None:
      raise HTTPException(422, detail="session_id required for SESSION scope")

  3. Confusing return structure in MemoryQdrantStore.get() — memory_qdrant.py:623-624:
  payload.update({"id": point.id, "score": None, "payload": dict(point.payload or {})})
  return payload
  Mutates payload dict and nests payload["payload"] — a copy of itself. This works with record_from_qdrant_payload but is confusing. Better to build a clean dict explicitly, matching scroll() and search() patterns:
  return {
      "record_id": (point.payload or {}).get("record_id", str(point.id)),
      "content": (point.payload or {}).get("content", ""),
      "score": None,
      "agent_id": (point.payload or {}).get("agent_id", ""),
      ...
      "payload": dict(point.payload or {}),
  }

  4. Redundant round-trip in delete() — memory_qdrant.py:687-689:
  existing = await self.get(record_id)
  if existing is None:
      return False
  Adds a retrieve call before every delete. But MemoryService.delete() already checks existence via PG (source of truth) — if PG returns deleted=False, the Qdrant delete is never reached. The double-check at the Qdrant level is     
  redundant and adds latency. Recommendation: keep a simple fire-and-forget delete, return bool based on PG (as MemoryService already does).

  ---
  WARNING

  5. No upper bound on offset — routes/memory.py:397:
  offset: int = Query(0, ge=0),
  scroll() in Qdrant over-fetches limit + offset points. With offset=100000, that loads 100,200 points into memory. Add le=10000 or a similar cap.

  6. No size limit on metadata — routes/memory.py:188:
  metadata: dict[str, Any] = Field(default_factory=dict)
  Without constraints, a client can send megabytes of metadata. Recommendation: add a validator for max keys count or max serialized size.

  7. get_current_user change is a separate concern — auth/dependencies.py:69:
  settings: Settings = request.app.state.settings
  Replacing Depends() with app.state.settings is the correct fix (previously each request instantiated a new Settings()). But this is an auth dependency fix unrelated to the memory API. Should be a separate commit for clean history.

  8. get_memory_service not part of lifecycle cleanup — dependencies.py:62-133. The AsyncEngine, RedisStore, and AsyncQdrantClient instances created here are never closed on shutdown. Lifespan is unaware of memory_services. Not a   
  blocker for this PR, but tech debt — worth a TODO comment.

  ---
  SUGGESTION

  9. MemoryRecordResponse and _record_to_response — consider adding model_config = ConfigDict(from_attributes=True) and using MemoryRecordResponse.model_validate(record, from_attributes=True) instead of manual field-by-field        
  mapping.

  10. serde.py adds hit.get("id") as a fallback in ID resolution — good addition for the get() response shape, but deserves an inline comment explaining why (for retrieve() results where record_id lives in payload, not at top       
  level).

PR #83, opened 2026-04-20
feat(MTRNIX-304): freshness worker for agent memory (Phase A)

godedok:
## Summary

Ports the metamemory freshness-worker concept onto metatron-core's `MemoryRecord` surface (agent memory only — Phase A). Five-stage bounded-loop pipeline (Linker → Reconciler → FreshnessMonitor → Curator → DecisionEngine) driven by a per-workspace Redis queue and a standalone worker process (`python -m metatron.memory.freshness`). DecisionEngine reuses `llm/provider.py`; `prometheus_client` is optional (gated `try/except ImportError`). Feature flag `METATRON_FRESHNESS_ENABLED=false` by default — zero behavioural change until flipped.

Follow-ups already split off (same sprint): **MTRNIX-313** (KB Phase B), **MTRNIX-314** (MCP `status` filter + review queue tools). Both blocked by this one.

- Spec: `docs/superpowers/specs/2026-04-20-freshness-worker-agent-memory-design.md`
- Plan: `docs/superpowers/plans/2026-04-20-freshness-worker-agent-memory.md`
- Confluence: [Freshness Workers in Metatron](https://mtrnix.atlassian.net/wiki/spaces/MTRNIX/pages/14745601)

## Changes

- `MemoryRecord` extended with 7 lifecycle fields (`status`, `freshness_score`, `superseded_by`, `valid_from/until`, `evidence_count`, `verification_state`). Backwards-compat defaults — existing rows land as `ACTIVE`.
- Alembic 016 — lifecycle columns on `memory_records` + new `review_entries` and `machine_events` tables.
- `storage/redis.py` gets queue + Lua-scripted lock primitives (LPUSH / BRPOP batch / `SET NX EX` lock with token-guarded release + heartbeat).
- `storage/memory_freshness_pg.py` — dedicated store for `review_entries` + `machine_events`.
- New module `src/metatron/memory/freshness/` — coordination, producer, 5 stages, DecisionEngine, worker, metrics, `__main__`.
- 4 new event constants in `core/events.py`: `FRESHNESS_JOB_ENQUEUED/PROCESSED`, `FRESHNESS_DECISION_APPLIED`, `FRESHNESS_REVIEW_CREATED`.
- Producer hook wired into `MemoryService.save / cache_session / delete` (no-op when flag off).
- 14 new `METATRON_FRESHNESS_*` env vars in `core/config.py`.
- `docker-compose.yml` — optional `freshness-worker` service under `--profile freshness`.

46 files changed, +5580 / −36. 13 per-task commits plus 2 upfront doc commits (spec + plan).

## Concerns flagged for reviewer

1. **Neo4j driver singleton** has no invalidation hook if a workspace is deleted while the worker runs long. Acceptable Phase A; revisit in Phase B.
2. **`FRESHNESS_REVIEW_CREATED` EventBus emission** — wired in Reconciler only; low-confidence-DecisionEngine path is left to the API layer (per spec). Will be completed in MTRNIX-314.
3. **First-run bulk STALE transition.** 30-day `stale_after_days` is compared against `updated_at`. Existing rows post-migration carry `status='active'` but may be >30d old — when we flip `FRESHNESS_ENABLED=true` in prod, Monitor can mass-transition them. By design, but consider a conservative first-run cap. Flag is off today; we have time to decide before flipping.

## Test plan

- [x] `make lint` clean on new / touched files (ruff check + format)
- [x] `make typecheck` — mypy strict zero errors on new code (repo baseline unchanged)
- [x] `make test` — 1600 pass / 84 fail / 22 errors. Baseline on `eb1abc4`: 1524 pass / same 84 fail / same 22 errors. **Zero regressions**; +76 new passing tests.
- [x] `pytest tests/unit/memory tests/unit/storage/test_memory_freshness_pg.py tests/unit/storage/test_memory_postgres_lifecycle.py tests/unit/storage/test_redis_coordination.py` — 219/219 green.
- [x] Integration: `test_enqueue_dequeue_processes_single_job` + `test_near_duplicate_creates_review_entry` against live PG/Qdrant/Redis/Neo4j.
- [x] Alembic `upgrade 016` + `downgrade 015 → upgrade 016` round-trip.
- [x] Smoke: `METATRON_FRESHNESS_ENABLED=false python -m metatron.memory.freshness` exits immediately.
- [ ] Reviewer: `make test-all` (integration with live services).
- [ ] Reviewer: backward-compat spot-check of `memory_store` / `memory_search` / `memory_batch_store` MCP tools with flag off.

## Coordination

`core/events.py` adds 4 new event constants — not breaking, but the enterprise plugin may want to subscribe to `FRESHNESS_DECISION_APPLIED` (audit) and `FRESHNESS_REVIEW_CREATED` (future Control Center review queue). Heads-up to the enterprise maintainer before merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 ### Added
+- feat: freshness worker for agent memory (Phase A, MTRNIX-304) — 5-stage bounded-loop pipeline (Linker → Reconciler → FreshnessMonitor → Curator → DecisionEngine), per-workspace Redis queue, standalone worker process. Feature-flagged via `METATRON_FRESHNESS_ENABLED` (default false). Adds 7 lifecycle fields to `memory_records` + `review_entries` + `machine_events` tables (migration 016). Follow-ups: MTRNIX-313 (KB Phase B), MTRNIX-314 (MCP status filter + review queue), MTRNIX-316 (queue reliability pre-prod gate).
 - MCP tools: `metatron_memory_search`, `metatron_memory_store`, `metatron_memory_delete` for agent memory CRUD via MCP (MTRNIX-303).
 - MCP tool: `metatron_search_fast` — low-latency document lookup (dense + optional metadata, P50 <800 ms target) (MTRNIX-303).
 - `fast_search()` function in `retrieval/search.py` — bypasses reranker / HyDE / query expansion / graph / LLM answer generation.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -38,6 +38,7 @@ Direct commands:
 python -m metatron          # API + legacy channel bots (Telegram/Discord/Slack) — channels scheduled for removal
 python -m metatron.api.app  # API only — preferred entry for new deployments
 python -m metatron.mcp      # MCP server (stdio/streamable-http) — for Hermes / Cursor / Claude Desktop
+python -m metatron.memory.freshness  # run freshness worker (requires METATRON_FRESHNESS_ENABLED=true)
 pytest tests/unit/test_search.py::test_name -v  # single test
 ```
 
@@ -105,6 +106,8 @@ src/metatron/
 ├── memory/                    # L3 — service.py (MemoryService orchestration, PG source of truth),
 │                              #   search.py (hybrid MemorySearchService), serde.py (Qdrant payload deserializer)
 │                              #   First-class new module (WS1). Assertion lifecycle layer planned on top.
+│   └── freshness/             # [MTRNIX-304] Linker, Reconciler, FreshnessMonitor, Curator, DecisionEngine
+│                              #   + coordination + worker; feature-flagged, SLM-backed
 ├── benchmarker/               # L2 — [OPTIONAL] api/, db/, schemas/, services/metrics/ — dev eval tool
 └── scripts/                   # graph_audit.py, run_eval.py, grid_search_weights.py,
                                # graph_rebuild.py, graph_process.py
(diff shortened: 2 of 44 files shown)


pahomm:
[WARNING] rpop_batch → potential job loss on worker crash

  CoordinationStore.dequeue_batch() uses Lua to atomically pull N jobs out of the queue (see redis.py:23-33 and coordination.py:89). If the worker crashes inside _process_job for job 5 out of 20, the remaining 15 have already been removed from Redis and   
  are not reprocessed anywhere — silent data loss.

  For Phase A with FRESHNESS_ENABLED=false this is acceptable, but once the feature is turned on, this becomes a risk. Future options: (a) BRPOPLPUSH into a processing-list with a reclaim-job; (b) reduce the batch size to 1. At a minimum — record this as a
   known limitation in docs/superpowers/specs/ before flipping the flag in prod.

  [WARNING] freshness_llm_api_base_url / _api_key are silently ignored for the default provider

  build_default_decision_engine() (decision_engine.py:204) passes api_url/api_key into create_provider(**kwargs). But _settings_for_provider (llm/provider.py:25-39) returns only {"model": ...} for ollama — api_url/api_key end up in **kwargs of
  OllamaProvider.__init__, which does not see them (host is taken from the OLLAMA_LLM_HOST/OLLAMA_HOST env vars). Only CustomProvider reads api_url.

  As a result, for METATRON_FRESHNESS_LLM_API_BASE_URL to actually take effect, the user must also set METATRON_FRESHNESS_LLM_PROVIDER=custom. This diverges from the design spec ("...read by a small factory that builds a dedicated provider independent of  
  the main chat LLM"). Options: (a) document the requirement in CLAUDE.md/HERMES; (b) in build_default_decision_engine, force provider_name="custom" when api_base_url is set; (c) teach OllamaProvider to read host from kwargs.

  [SUGGESTION] N+1 UPDATE in apply_decision

  decision_engine.py:243-248 — one separate await pg_store.update_lifecycle(..., append_tag=tag) per tag, each in its own transaction (engine.begin()). For 5 tags that's 5 roundtrips. Better to accept append_tags: list[str] and do it in a single UPDATE via
   tags || :tag_array with dedup.

  [SUGGESTION] Checkpoints are written but never read for idempotence

  Every stage calls write_checkpoint at the end, but nobody checks read_checkpoint before doing work. The design spec says "per-stage last-completion marker, used for idempotent re-runs." Either add a check at the start of each stage (and if
  stage:{record_id} == expected → skip), or honestly drop the checkpoint-writing code and keep only MachineEvent as the audit trail.

  [SUGGESTION] Neo4j writes — thread-pool pressure

  Linker.run (linker.py:132-141) issues asyncio.to_thread(link_memory_items, ...) in a loop over related records (up to 20 edges). Each call holds a separate sync driver session. With 20 workers × 20 edges that's 400 concurrent thread-pool workers at once.
   It would be enough to batch into a single Cypher UNWIND $edges AS e MATCH ... MERGE ....

  [SUGGESTION] scan_keys without workspace-prefix safeguard

  CoordinationStore.list_active_workspaces calls scan_keys("freshness:queue:*"). If Redis is shared between environments (dev/staging on a single instance), this can surface workspace ids from other environments. Not a blocker — prod has its own Redis —   
  but collisions are possible in the docker-compose dev environment.

  [NIT] Dead import

  reconciler.py:27 imports FRESHNESS_REVIEW_CREATED at module top level and only uses it when event_bus is not None. In this PR the event_bus is never passed through (neither in _build_worker nor in the e2e test), so the event is effectively never emitted.
   Either wire the bus into _build_worker, or drop the branch until MTRNIX-314 needs a subscriber.

  [NIT] settings = get_settings() is called from many hot paths

  In _process_job, _run_loop, enqueue_if_enabled, apply_decision, build_default_decision_engine — each time freshly. Pydantic caches it, but stylistically it is cleaner to thread settings through DI once.

PR #84, opened 2026-04-21
feat(MTRNIX-270): agent registry backend — CRUD, lifecycle, versioned…

pahomm:
… config

New L3 module `metatron.agents` — first-class identity primitive for external agent runtimes (Hermes, Cursor, Claude Desktop). REST API at /api/v1/agents/*: CRUD, soft-delete, lifecycle flag (start/stop/pause), full version history.

Scope:
- migration 016 — agents + agent_config_versions tables
- models / service / persistence following the memory/ pattern
- 9 endpoints, reads gated by require_viewer, writes by require_editor
- PUT atomically bumps config_version and appends a full snapshot row
- memory_bindings and budget stored as opaque JSONB (enforcement deferred)
- memory_records.agent_id stays a soft-reference (no FK) — Hermes can keep writing memory without prior registration
- 45 new unit tests, zero regressions, docker build verified

Out of scope (tracked separately): company/team hierarchy, 5-role RBAC, MCP exposure of agent registry, budget/bindings enforcement — all deferred to follow-up tickets or the future CC plugin.


godedok:
## Review — MTRNIX-270: Agent Registry backend

Architecturally clean work: new `agents/` module mirrors the `memory/` layout, DI lives on `app.state`, `PUT` is transactional (`SELECT … FOR UPDATE` → UPDATE + version-row INSERT in one tx), workspace-isolation on `list_versions` is enforced via JOIN, and typed `MetatronError` subclasses map to proper HTTP codes at the route boundary. The `docs/DECISIONS.md` entry resolves the long-standing "core vs CC repo" ambiguity for agent identity in a defensible way.

**Recommendation: do not merge yet** — the two BLOCKER-level items below are documentation-vs-code contract violations and fix in a small follow-up commit on the same feature branch.

### 🔴 BLOCKER

**1. `list_agents` returns archived rows by default — contradicts the documented contract.**
`src/metatron/agents/.claude/CLAUDE.md`:
> "Soft-delete via ARCHIVED — no rows are removed; the list endpoint defaults to non-archived."

But `persistence.py:217-242` and `service.py:129-144` only apply the status filter when `status is not None`. So `DELETE /api/v1/agents/{id}` flips status to `ARCHIVED`, and the agent still shows up in `GET /api/v1/agents` with no filter. Either add an "exclude archived" default to `list_records` (and provide an explicit opt-in like `include_archived=true`) or correct the doc to match current behavior.

### 🟡 WARNING

**2. Unique index `uq_agents_workspace_name` blocks name reuse after soft-delete.**
`migrations/versions/016_agents.py:92-97` creates a non-partial unique index on `(workspace_id, name)`. After `DELETE`, the archived row still occupies the name — creating a fresh agent with the same name returns 409. Either switch to a partial index (`WHERE status != 'archived'`) or document this deliberately as an audit-integrity constraint. Current behavior is surprising UX.

**3. `TestWorkspaceIsolation` (`tests/unit/test_agents_routes.py:421-438`) is misnamed.**
The two tests only assert that `created_by` / `changed_by` are taken from the authenticated user. There is no actual cross-workspace check (two services with different `workspace_id`s, verifying data separation at the persistence layer). Either rename to `TestAuditFields`, or add a real cross-workspace test — e.g. service A creates agent X in ws-A, service B bound to ws-B must not be able to `get`/`update`/`list_versions` on X.

**4. No integration tests against PostgreSQL for `AgentPersistence`.**
All service-layer tests use `AsyncMock` for the repo. The `FOR UPDATE` semantics, `IntegrityError → _AgentNameConflictError` mapping, and the JOIN-based workspace enforcement in `list_versions` are never exercised against a real database. At minimum, the race path stubbed by `test_race_disappearance_raises_not_found` should have a real concurrent-tx counterpart. The existing `memory/` module has such tests — pattern is already in the repo.

**5. `get_workspace_id` (`api/dependencies.py:57-69`) silently collapses `workspace_ids=["*"]` to `default_workspace_id`.**
Admin tokens with wildcard workspace access land on the default workspace by surprise — so an admin tool creating an agent ends up in the wrong tenant. Inherited from `memory/`, not a regression from this PR, but worth a follow-up for the 5-role RBAC work: an admin calling `/api/v1/agents` should probably be required to name the workspace explicitly.

### 🟢 SUGGESTION

**6. Empty-body `PUT` is only blocked by the request schema.** `UpdateAgentRequest._validate` rejects bodies with no fields (422), but `service.update_agent(agent_id, changed_by=…)` with no fields happily bumps `config_version` (confirmed by `test_empty_update_only_bumps_version`). When a second entry point lands (MCP tool, future CC plugin direct call), the guard is bypassed. Consider duplicating the "at least one field" check in the service.

**7. `ORDER BY created_at DESC, id ASC`** ties by `uuid4` hex, which is deterministic but visually random. Fine for MVP; for a future admin UI, `updated_at DESC` as the secondary key is more human-friendly.

**8. `memory_records.agent_id` stays a free string without a FK** — deliberate per DECISIONS.md §4, and the right call for Hermes right now. But that means "show me all memory belonging to agent X" cannot be a JOIN. When CC Memory Governance arrives, a backfill / normalization pass will be needed. Worth a follow-up ticket so it doesn't get lost.

**9. `AgentResponse` always returns `memory_bindings` and `budget` to any viewer+.** When CC adds budget enforcement these may become sensitive. MVP is fine; flag as a candidate for field-level masking during the 5-role rework.

### ⚙️ PROCEDURAL

**10. Commit `6219aa84` carries a `Co-Authored-By: Claude …` trailer.** Per project convention, squash-merge should drop it — worth checking the merge message explicitly before merging so the trailer doesn't end up on `develop`.

---

### Summary

- ✅ Architecture, DI pattern, transactional PUT, workspace isolation at SQL layer — solid.
- ✅ `docs/DECISIONS.md` entry resolves the "core vs CC repo" question properly.
- 🔴 Items #1 and #2 are real doc-vs-code contract violations and should be fixed before merge (small follow-up commit on `feature/MTRNIX-270`).
- 🟡 Items #3–#5 can land as separate follow-ups but the real workspace-isolation test (#3) is strongly recommended on this PR given how central isolation is to the product.

PR #99, opened 2026-04-27
feat(WS4 S6): agent activity logging — event table + EventBus emissio…

pahomm:
…n + API

Introduces observability-foundation activity log:
- agent_activity_log (append-only) + 2 indexes (migration 019)
- 8 new EventBus constants (memory_promoted, document_accessed, tool_called, agent_created/updated/status_changed/deleted, error_occurred)
- activity/ L3 module: ActivityLogger (EventBus subscriber) + ActivityService (read facade) + current_agent_id ContextVar
- AgentIdContextMiddleware registered in both create_app() and mcp/server.py run_http() for mounted + standalone MCP transport
- Emission hooks: memory/service.py (save/delete/promote/reset with N+umbrella), agents/service.py (create/update/status_changed/deleted, delete bypasses status_changed), retrieval/search.py (correlation_id + query.processed + per-channel document.accessed), mcp/server.py (tool wrapper with 8 KiB arg cap via monkey-patch of mcp.tool decorator)
- GET /api/v1/agents/{id}/activity + /activity/summary (require_viewer)
- METATRON_ACTIVITY_LOG_ENABLED feature flag (default true)
- api/middleware.py converted to api/middleware/ package (R100 rename)

Review fixes (verified end-to-end on live PG via integration smoke):
- B1: lazy-resolve _ACTIVITY_BUS_GETTER at every wrapper invocation so tools decorated at import time still pick up the bus installed by create_app(). Regression test: test_bus_getter_resolved_lazily.
- B2: get_memory_service (REST) and mcp/tools/_memory_deps (MCP) now wire event_bus into MemoryService — without this no memory.* events were written. Added get_activity_bus() accessor on mcp/server.py for the MCP side. Regression test: test_dependencies_memory_event_bus.
- I3: tool wrapper binds agent_id into the contextvar before calling the handler, so hybrid_search_and_answer reads the same value the wrapper observed. Regression test: test_wrapper_binds_agent_id_to_contextvar.
- I4: chat.py / openai_compat.py / mcp/tools/search.py now pass source="rest"|"oai_compat"|"mcp" — query.processed payload no longer always reports "rest".
- I8: added test_missing_workspace_id_is_skipped covering the symmetric drop path next to skipped_no_agent_id.

Tests: +66 unit tests, no regressions vs develop baseline; integration smoke passes against live PG (memory.created + agent.created + agent.deleted all observed).

Spec: docs/superpowers/specs/2026-04-23-agent-activity-logging-design.md
Plan: docs/superpowers/plans/2026-04-23-agent-activity-logging.md


godedok (commented):
## Code Review — feat(WS4 S6): Agent Activity Logging

Overall the PR is well-structured: L3 `activity/` module is cleanly separated, the review fixes (B1–I8) are addressed with regression tests, and the EventBus subscription pattern follows existing conventions. Two blockers need to be resolved before merge.

---

### BLOCKERS

**[B1] Workspace isolation not explicitly verified in activity endpoints**
`api/routes/agents.py:789–833`

`get_agent_activity()` checks agent existence via `reg_service.get_agent(agent_id)`, but never explicitly asserts `agent.workspace_id == resolved_workspace_id` before querying the activity store. If there is ever a regression in registry isolation, activity rows from a foreign agent could leak.

```python
agent = await reg_service.get_agent(agent_id)
if agent.workspace_id != workspace_id:
    raise HTTPException(status_code=404)
```

---

**[B2] Activity endpoints not guarded when `METATRON_ACTIVITY_LOG_ENABLED=false`**
`api/routes/agents.py`, `api/dependencies.py`

When the flag is off, `ActivityStore` is never instantiated, but the routes are still registered. The dependency will resolve to `None` and the first attribute access will raise `AttributeError` at runtime. The endpoint should return `503 Service Unavailable` (or `404`) explicitly.

Side-effect: with the flag off, EventBus still emits activity events — the subscriber simply isn't registered so they are silently dropped. If this is intentional (bus overhead is negligible), it should be documented; otherwise gate the emission call-sites on the flag too.

---

### WARNINGS

**[W1] Argument truncation (8 KiB cap) is silent**
`mcp/server.py:1024–1038`

`_project_arguments()` sets `arguments_truncated=True` but emits no log line. Tool authors will never know their payload was capped. Add a `logger.warning("activity_log.arguments_truncated", tool=..., original_size=...)` when truncation occurs.

---

**[W2] `event_bus=None` drops events without any warning**
`agents/service.py:451`, `memory/service.py:1221`

Both services silently skip emission when `self._event_bus is None`. A misconfigured deployment (bus not wired) would lose all events with no operator signal. A one-time `logger.warning("event_bus_not_wired", service=...)` on first skipped emission would surface this.

---

**[W3] `document_ids` cap in search emission is silent**
`retrieval/search.py:1408`

```python
"document_ids": list(doc_ids)[:50],
```

Hard cap at 50 with no log — same pattern as W1. Add `activity_log.document_ids_capped` warning when the list is truncated.

---

**[W4] `_ACTIVITY_BUS_GETTER` global has no documented single-call contract**
`mcp/server.py:998–1011`

The lazy-resolve pattern is correct, but `set_activity_bus_getter()` gives no signal that calling it a second time (e.g., in tests or hot-reload) would silently replace the getter. An `assert` or a one-line docstring stating the expected call contract would prevent future confusion.

---

### SUGGESTIONS

**[S1] Composite index would help `summary_for_agent` at scale**
`migrations/versions/019_agent_activity_log.py`

The existing `ix_activity_agent (workspace_id, agent_id, created_at DESC)` covers listing queries. The summary query additionally groups by `event_type` — a partial composite index on `(workspace_id, agent_id, event_type, created_at DESC)` would avoid a seq scan on high-volume agents.

**[S2] PII risk in raw `event_data` JSONB**
`activity/logger.py:311`

Event data is stored unfiltered. If a tool receives credentials or PII as arguments, they will end up in the activity log. No code change needed now, but `HERMES_INTEGRATION.md` should document that tools must not pass sensitive data as arguments.

**[S3] ContextVar reset in `finally` and unawaited tasks**
`mcp/server.py:1053–1075`

The wrapper resets `current_agent_id` in `finally`. Any tool that spawns a fire-and-forget background task will see a reset context var in that task. No such tools exist today, but a short warning comment next to the `finally` block would prevent a hard-to-debug issue later.

**[S4] Zero-fill boundary in `summary_for_agent` is implicit**
`storage/activity_pg.py:1620–1628`

```python
end_day = (until.astimezone(UTC) - timedelta(microseconds=1)).date()
```

Logic is correct, but it is non-obvious that `until=2026-04-23T00:00Z` excludes April 23rd. One comment line would save the next reader.

---

### Test gaps

- No test for `503`/`404` when `METATRON_ACTIVITY_LOG_ENABLED=false` and an activity endpoint is called.
- No concurrent ContextVar isolation stress test (N parallel requests with distinct `X-Agent-Id` headers).
- No test for `event_data` payload near the PostgreSQL JSONB limit.

---

### Summary

| | Count |
|---|---|
| BLOCKER | 2 |
| WARNING | 4 |
| SUGGESTION | 4 |
| Test gaps | 3 |

**Do not merge** until B1 and B2 are resolved. W1–W4 should ideally be addressed in this PR; the rest can be handled in follow-up.

PR #103, opened 2026-05-04
feat(MTRNIX-272): memory snapshot/restore/diff (WS1 S4-5)

pahomm:
MemorySnapshotService — JSONL+gzip+SHA256 backup/restore for agent memory. Atomic file write (tmp+rename for both gzip and sidecar), pre-reset/pre-restore auto-snapshots, transactional PG replace via MemoryPostgresStore.replace_for_agent, best-effort Qdrant + Neo4j re-population. REST endpoints under /api/v1/agents/{id}/{reset,snapshots} and /api/v1/snapshots/{id}/restore + /diff. Same-agent diff only. RuntimeError on >10k overflow -> 413; SnapshotCorruptError -> 422; reset failure after pre_reset snapshot returns 500 with snapshot_id in detail for operator recovery. Storage path is relative to METATRON_SNAPSHOT_DIR.

Also: add /logs/ to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,9 @@ docs/superpowers/
 # Data
 /data/
 
+# Logs
+/logs/
+
 # Testing
 .pytest_cache/
 eval_results/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,20 @@
 ## [Unreleased]
 
 ### Added
+- feat: Memory snapshot/restore/diff for agent memory (MTRNIX-272, WS1 S4-5).
+  `MemorySnapshotService` — JSONL+gzip+SHA256 backups with atomic file write
+  (tmp+rename for both gzip and sidecar). Pre-reset and pre-restore auto-snapshots
+  guarantee an undo point before any destructive operation. Restore uses a single
+  `MemoryPostgresStore.replace_for_agent` transaction (DELETE+INSERT) with best-effort
+  Qdrant+Neo4j repopulation (PG remains source of truth). New REST endpoints:
+  `POST /api/v1/agents/{id}/reset` (auto pre_reset snapshot → wipe, 413/422/500-with-snapshot-id),
+  `POST /api/v1/agents/{id}/snapshots` (manual snapshot, 201),
+  `GET /api/v1/agents/{id}/snapshots` (list newest-first),
+  `POST /api/v1/snapshots/{id}/restore` (checksum verify → pre_restore snapshot → transactional replace),
+  `GET /api/v1/snapshots/diff` (same-agent comparison, `key=source|content_hash`).
+  New events: `MEMORY_SNAPSHOT_CREATED` (`snapshot_id`, `trigger`, `record_count`),
+  `MEMORY_RESTORED` (`snapshot_id`, `record_count`, `pre_restore_snapshot_id`).
+  New env vars: `METATRON_SNAPSHOT_DIR` (./data/snapshots), `METATRON_SNAPSHOT_MAX_FILE_BYTES` (256 MiB).
 - docs: strategy ADR `docs/adr/2026-04-25-metatron-strategy.md` —
   authoritative snapshot of architectural decisions and pre-pilot plan
   (Memory Quality Layer with `kind=fact|preference|pinned`, Agent
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -278,6 +278,8 @@ Graph extraction is decoupled from sync (process_all_unsynced_graphs, graph-proc
 - METATRON_FRESHNESS_KB_SEARCH_FILTER_ENABLED (false) — retrieval-side ARCHIVED/SUPERSEDED filter pushdown; when on, recall channels combine the filter with `access_filter` via `_combine_filters`
 - METATRON_FRESHNESS_WEIGHT (0.0) — scoring weight for the `freshness` signal in `compute_signal_score`; default 0.0 keeps the formula numerically identical to Phase A
 - METATRON_FRESHNESS_KB_STALE_AFTER_DAYS (90) — KB stale threshold in days (higher than memory's 30 because KB documents age more slowly)
+- METATRON_SNAPSHOT_DIR (./data/snapshots) — root directory for memory snapshot files (MTRNIX-272)
+- METATRON_SNAPSHOT_MAX_FILE_BYTES (268435456) — per-file size cap (256 MiB) for snapshot exports; exceeded → 413 (MTRNIX-272)
 - See core/config.py for full list
 
 ## External Agent Integration Surfaces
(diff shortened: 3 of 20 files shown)


godedok:
## Review — MTRNIX-272 memory snapshot/restore/diff

Solid PR overall: the design is sound (PG as source of truth, atomic file writes, SHA-256 with sidecar, manifest cross-check, transactional `replace_for_agent`, pre-reset / pre-restore auto-snapshots, workspace-bound service). Tests have good coverage of the happy paths and the corruption / overflow / RBAC edge cases. Below are the things I'd want addressed before merge.

### Blockers

**1. `POST /api/v1/snapshots/{id}/restore` does not catch `RuntimeError` from the auto pre-restore snapshot.**
`MemorySnapshotService.restore` calls `self.create(..., trigger=PRE_RESTORE)` internally, and `create` raises `RuntimeError` when the agent currently holds more than 10 000 records (the overflow guard at `memory/snapshot.py`). The route only catches `MemoryNotFoundError` / `SnapshotCorruptError`, so an agent that grew past the cap since the snapshot was taken will surface a generic `500 Internal Server Error` with no operator-actionable detail. Either catch `RuntimeError` → 413 here as well (mirror `agents.py:reset_agent_memory` and `create_agent_snapshot`), or — better — introduce a dedicated `SnapshotOverflowError` and catch that everywhere.

**2. Catching bare `RuntimeError` for the overflow case is too broad.**
`agents.py` has `except RuntimeError as exc: raise HTTPException(413, ...)`. Any unrelated `RuntimeError` propagating from the snapshot service (e.g. an IO error wrapped as RuntimeError, or a future bug) will be silently mapped to 413 "payload too large". Replace the `raise RuntimeError(...)` in `MemorySnapshotService.create` with a specific exception type and catch that instead.

### Warnings

**3. Inconsistent HTTP semantics for size-related failures.**
- `>10 000 records` → `RuntimeError` → **413**.
- File exceeds `max_file_bytes` → `SnapshotCorruptError` → **422**.

Both are "payload too large" conditions; using two different status codes (and labelling one as "corrupt") is confusing. Suggest unifying both under 413 with a dedicated overflow exception, and reserving `SnapshotCorruptError` / 422 strictly for integrity failures (checksum mismatch, manifest mismatch, malformed JSON).

**4. Restore does per-id Qdrant delete but agent-wide Neo4j delete — asymmetric and partially divergent from `MemoryService.reset`.**
In `MemorySnapshotService.restore`:
- Qdrant cleanup iterates `deleted_ids` returned by `replace_for_agent` (per-id delete).
- Neo4j cleanup calls `_graph_delete_agent_memories(workspace_id, agent_id)` (wholesale).

`MemoryService.reset` explicitly comments that it uses per-id deletes "to avoid the over-delete bug of `delete_by_agent`". The restore path uses `delete_by_agent` for the graph anyway. Pick one model and apply it consistently — and if `delete_by_agent` is genuinely buggy (per the existing comment), the restore path inherits that bug. Worth a quick check.

A second related issue: per-id Qdrant delete will leave orphans behind if Qdrant somehow holds points whose ids aren't in PG (e.g. from a previously-aborted write). For a "replace everything for this agent" operation, an agent-scoped wipe is more defensive — provided the underlying bug noted in `MemoryService.reset` doesn't apply here.

**5. Restore is fully synchronous and re-embeds every record.**
For a 10 k-record agent the route awaits N sequential `qdrant.upsert` calls (each potentially round-tripping to the embedding provider) plus N `to_thread` Neo4j writes. Realistic HTTP timeouts will fire well before completion. Either batch the upserts (`qdrant.upsert_batch` if available) or move the downstream repopulation to a background task and return 202 + a job id. At minimum, document the practical record-count ceiling in the route docstring.

**6. Documentation/implementation drift on Qdrant restore.**
The plan in `docs/superpowers/plans/2026-05-02-ws1-snapshot-restore-diff.md` (step 6 of `restore`) says: `qdrant_store.delete_by_agent(agent_id) then qdrant_store.upsert(record) per record`. The implementation does per-id delete instead. Either update the plan or align the implementation; right now the two disagree.

### Suggestions

**7. `_diff_key` docstring lists only `src::` and `id::` prefixes** but the function actually emits three (`hash::`, `src::`, `id::`). Update the docstring to mention `hash::` for `key=CONTENT_HASH`.

**8. `tests/unit/memory/test_snapshot_service.py::TestCreate::test_workspace_mismatch_raises` is misnamed** — it asserts `agent_id is required`, not a workspace mismatch. Rename to `test_create_rejects_empty_agent_id` (and consider adding a real workspace-mismatch test that exercises `replace_for_agent`'s validation path end-to-end).

**9. `CreateSnapshotRequest.model_config = ConfigDict(strict=False)` is the pydantic v2 default** — can be removed without behaviour change.

**10. No per-agent locking around restore.** Two concurrent `POST /snapshots/{id}/restore` for the same agent could interleave. PG is protected by the transaction, but Qdrant/Neo4j repopulation is not, so the surviving downstream state can be partially from each restore until the next write reconciles. Acceptable for v1, but worth either documenting or adding a per-(workspace, agent) Redis lock.

**11. `embedding` is intentionally excluded from the snapshot** (re-embed on restore) — fine, and matches the plan. Worth a single sentence in `snapshot.py` module docstring stating this explicitly so future readers don't think it's an oversight.

**12. Orphaned-file cleanup on crash.**
If the process dies between the gzip rename and `pg_store.save_snapshot`, an orphan `.jsonl.gz` (and possibly `.sha256`) remain on disk with no PG row pointing to them. The cleanup happens only inside the `try/except`. A simple startup sweep ("delete files under `snapshot_dir/{ws}/{ag}/` not present in `memory_snapshots`") would close this — listed as out-of-scope already, just flagging that retention/cleanup follow-up should also include orphan reaping, not just expired-snapshot pruning.

### Things I checked and liked

- SHA-256 over the on-disk gzip bytes (not plaintext) — file integrity verifiable without re-parsing.
- Sidecar verification cross-checks PG `content_hash` against the sidecar — defends against a synchronized PG-row + file tamper.
- `replace_for_agent` validates each record's `workspace_id`/`agent_id` before opening the transaction — fail-fast and avoids partial deletes.
- Service derives the file path from canonical layout, not from the `storage_path` column — prevents path-traversal via tampered PG row.
- Manifest sanity-check against the snapshot row catches a swapped-file attack.
- `os.replace` for both gzip and sidecar (atomic on POSIX), `mtime=0` in `GzipFile` for deterministic output.
- Event payload shapes for `MEMORY_SNAPSHOT_CREATED` / `MEMORY_RESTORED` match the comment in `core/events.py`.
- File-rollback-on-PG-failure covered by tests.
- RBAC: editor+ for write paths, viewer+ for reads — consistent with the rest of `/api/v1/memory`.

Overall: address the two blockers (route catches `RuntimeError` for restore, dedicated overflow exception) and the per-id-vs-agent-wide cleanup divergence (#4) before merge. Everything else is polish.

PR #107, opened 2026-05-07
Feature/mtrnix 277

pahomm:


godedok:
### Review — MTRNIX-277 (memory health endpoint)

Overall the change matches the ticket scope tightly: read-only `GET /api/v1/agents/{id}/memory/health`, two new columns (`last_accessed_at`, `content_simhash`) with partial indexes, simhash-based near-duplicate clustering off the loop, search-side touch hook, env-driven knobs, and a backfill script. RBAC (`require_viewer`) plus the explicit `agent.workspace_id != reg_service.workspace_id` defence-in-depth check is the right pattern. PG-only fan-out via `asyncio.gather` keeps the endpoint cheap, and the `asyncio.to_thread` offload for the O(N²) hamming compare is the right call. Test coverage hits the meaningful branches — hardcap skip, null-simhash counting, ACTIVE-only filtering, zero-fill timeseries, viewer/editor/anonymous matrix, cross-workspace 404, signed↔unsigned bigint round-trip.

A few things worth addressing before merge:

**1. Write amplification on the touch hook (medium).** `MemorySearchService.search` in `src/metatron/memory/search.py:253-263` schedules a `bulk_touch_last_accessed` task for every successful search, unconditionally and per-record, with no throttling or dedup. A hot agent doing N searches per minute returning K records each issues N updates touching N·K rows per minute — pure WAL/autovacuum load that wasn't there before. Recommendation: add a freshness predicate to the UPDATE so touches no-op when the row was already touched recently, e.g.

```sql
UPDATE memory_records SET last_accessed_at = NOW()
WHERE workspace_id = :ws AND agent_id = :agent_id AND id = ANY(:ids)
  AND (last_accessed_at IS NULL OR last_accessed_at < NOW() - INTERVAL '1 minute')
```

This keeps stale-detection semantics intact (1-minute resolution is more than enough for a 30-day staleness window) and collapses the per-search write storm to ~once-per-minute-per-record.

**2. Silent duplicate-detection skip at the hardcap (low).** In `src/metatron/memory/health.py:112-120`, when `total > _DUP_HARDCAP` (5000) the response carries `duplicate_ratio=0.0, duplicate_clusters_count=0` with only a `logger.warning`. From the API caller's perspective this is indistinguishable from "no duplicates found", which is the wrong inference for the largest agents — exactly the ones an operator would want to inspect. Suggest adding an explicit boolean to `AgentMemoryHealth` / `MemoryHealthResponse` (e.g. `duplicate_detection_skipped: bool` plus `duplicate_active_population: int`) so the W9 dashboard can render "Skipped, agent over 5k active records" instead of a misleading 0%.

**3. `MemoryRecord.content_simhash` default of `0` (minor).** `src/metatron/core/models.py:378` ships `content_simhash: int = 0` with a comment claiming `0 == not yet computed`, but the PG layer uses NULL for that, and `_simhash` can legitimately produce 0 for empty/whitespace content. The end-to-end safety net is `if s` in `_cluster_duplicates_sync` (treats 0 as missing), so it works in practice, but the type is ambiguous. Either typing it `int | None = None` (matches the NULL semantics in PG) or dropping the default and having `MemoryService.save` always compute before persistence would be clearer.

**4. `iter_simhashes_active` is not actually an iterator (nit).** `src/metatron/storage/memory_postgres.py:1006-1030` is named `iter_…` but materialises a full `list[tuple[str, int]]` in memory. With `_DUP_HARDCAP=5000` it's bounded, but the prefix is misleading next to the genuine streaming `iter_records_missing_simhash` two methods down. Consider `list_simhashes_active` or wire it through `yield` in batches.

**5. Health-snapshot consistency window (nit).** `_compute_duplicates` fans out `iter_simhashes_active` and `count_active_with_null_simhash` concurrently — a record created/updated between those two reads can show up in one but not the other, slightly skewing the `null_count` warning. Fine for an observability endpoint, just calling it out so it doesn't look like a bug later.

**6. Backfill script global mode.** `scripts/backfill_memory_simhash.py` walking every workspace with `workspace_id=None` is fine for the migration window, but on a multi-tenant DB with millions of rows the single global pass holds a transaction per batch with no concurrency. If we intend to run this on production scale, a `--workspace-id` per-tenant loop driven externally is safer than the single-pass default. Documenting that recommendation in the script's docstring would be enough.

Migration 021 itself looks clean — both columns nullable with proper NULL-aware partial indexes, balanced upgrade/downgrade, no destructive ops. Backward compat for existing rows (NULL `last_accessed_at` falling back to `created_at` in `count_unused`) is handled correctly.

None of the above are blockers in principle, but #1 and #2 affect production behaviour visibly enough that I'd want them addressed in this PR rather than as a follow-up.

PR #114, opened 2026-05-13
fix(MTRNIX-332): migrate sync cursor to PG + sub-minute Jira post-filter

pahomm:
The file-based SyncState was the source of the "every sync returns 1 fetched" trap on prod: the cursor file on disk silently failed to advance (write failure or multi-replica filesystem isolation) while PG's last_synced_at advanced correctly. Each sync then re-fetched the same issue because the JQL filter used a cursor that never moved past the issue's updated timestamp.

Migrate the incremental cursor to PostgreSQL connections.last_synced_at:

- _run_connection_sync receives last_synced_at as a kwarg from trigger_sync (parsed from the connection record's ISO string)
- File-write to .metatron/sync_state.json is removed from the success path (PG already stamps in update_connection_status)
- NULL cursor (freshly-created connection) yields since=None -> full fetch on first sync
- Failed syncs no longer advance the cursor: pass None as last_synced_at on non-success status so the SET clause is skipped. Without this guard, transient failures silently drop documents updated between the last good sync and the failure

Jira-side fix for sub-minute precision: JQL's date filter only supports HH:mm granularity, so a cursor at 22:09:40 formatted as "22:09" still matches issues updated 22:09:00-22:09:59. Add a post-filter using raw_issue.fields.updated (no second process_jira_issue pass) that drops anything whose actual updated <= since.

Admin /reindex now clears connections.last_synced_at = NULL in the same transaction as the raw_documents reset; legacy file-based SyncState is also cleared best-effort for rolling-deploy compatibility.

Tests cover the regression matrix: cursor read from PG, force_full bypass, NULL initial state, failed-sync does-not-advance, Jira post-filter equal/greater/since=None branches, admin reindex clears PG cursor.

Follow-ups (not in this PR):
- Remove SyncState class and test_incremental_sync.py after one release cycle
- Tighten JQL date format with explicit TZ (Atlassian profile-TZ ambiguity in the minute-precision string)

### DIFF
--- a/src/metatron/api/routes/admin.py
+++ b/src/metatron/api/routes/admin.py
@@ -9,9 +9,11 @@ from __future__ import annotations
 from typing import Any
 
 import structlog
-from fastapi import APIRouter, Header, HTTPException
+from fastapi import APIRouter, Header, HTTPException, Request
 from pydantic import BaseModel
 
+from metatron.core.config import Settings
+from metatron.storage.postgres import PostgresStore
 from metatron.storage.cleanup import (
     ALLOW_CLEANUP,
     CleanupError,
@@ -99,9 +101,18 @@ class ReindexResponse(BaseModel):
 
 @router.post("/reindex", response_model=ReindexResponse)
 async def trigger_reindex(
+    request: Request,
     x_confirm_reindex: str | None = Header(None),
 ) -> ReindexResponse:
     """Trigger full reindex: reset sync flags and clear Neo4j graph.
+
+    **GLOBAL operation — affects EVERY workspace in this deployment.** There
+    is no per-workspace scoping; ``connections.last_synced_at`` is NULL'd for
+    all rows and ``raw_documents`` qdrant/graph flags are reset across the
+    whole table. On a multi-tenant deployment this triggers a re-embed and
+    re-graph storm on every tenant simultaneously. The ``X-Confirm-Reindex:
+    yes`` header is the only safety net — operators must understand the
+    blast radius. Per-workspace reindex is a separate follow-up.
 
     Does NOT require ALLOW_CLEANUP. After calling this, trigger sync from UI
     to re-ingest all documents with current settings (e.g. SPLADE vectors).
@@ -128,31 +139,28 @@ async def trigger_reindex(
 
     # 1. Reset PG sync state — clear connections.last_synced_at AND
     # raw_documents.{qdrant,graph}_synced flags in a single transaction so
-    # we don't keep two parallel engines for adjacent UPDATEs (MTRNIX-332).
+    # we use one connection from the pool (MTRNIX-332).
     # Next sync starts from since=None (full fetch).
+    store: PostgresStore | None = getattr(request.app.state, "postgres", None)
+    if store is None:
+        settings: Settings = request.app.state.settings
+        store = PostgresStore(settings.postgres_dsn)
+        request.app.state.postgres = store
     try:
         from sqlalchemy import text
 
-        from metatron.core.config import Settings
-        from metatron.storage.postgres import PostgresStore
-
-        settings = Settings()
-        store = PostgresStore(settings.postgres_dsn)
-        try:
-            async with store._engine.begin() as conn:
-                await conn.execute(text("UPDATE connections SET last_synced_at = NULL"))
-                r = await conn.execute(
-                    text(
-                        "UPDATE raw_documents "
-                        "SET qdrant_synced = false, qdrant_synced_at = NULL, "
-                        "    graph_synced = false, graph_synced_at = NULL"
-                    )
+        async with store._engine.begin() as conn:
+            await conn.execute(text("UPDATE connections SET last_synced_at = NULL"))
+            r = await conn.execute(
+                text(
+                    "UPDATE raw_documents "
+                    "SET qdrant_synced = false, qdrant_synced_at = NULL, "
+                    "    graph_synced = false, graph_synced_at = NULL"
                 )
-                docs_reset = r.rowcount
-            sync_state_cleared = True
-            logger.info("admin.reindex.pg_reset_done", docs_reset=docs_reset)
-        finally:
-            await store.close()
+            )
+            docs_reset = r.rowcount
+        sync_state_cleared = True
+        logger.info("admin.reindex.pg_reset_done", docs_reset=docs_reset)
     except Exception as e:
         logger.warning("admin.reindex.pg_reset_error", error=str(e))
 
--- a/src/metatron/api/routes/admin.py
+++ b/src/metatron/api/routes/admin.py
@@ -107,10 +107,12 @@ async def trigger_reindex(
     to re-ingest all documents with current settings (e.g. SPLADE vectors).
 
     Steps:
-    1. Reset sync state (forces full fetch from connectors)
-    2. Reset qdrant_synced=false for all raw_documents
-    3. Reset graph_synced=false for all raw_documents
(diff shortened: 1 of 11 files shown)


godedok:
## Code review — MTRNIX-332

Solid fix for the core trap (file-based cursor diverging from PG). A few correctness and consistency issues worth addressing before merge.

### BLOCKER — Cursor stamped at `NOW()` instead of `fetch_started_at`

`src/metatron/api/routes/connections.py` (`_run_connection_sync`, finally block):
```python
last_synced_at=(datetime.now(UTC) if status in ("success", "partial") else None),
```

The cursor is captured **after** `connector.fetch` returns, ingestion runs, graph extraction runs, etc. Any source-side update that happens between `fetch_started_at` and this `datetime.now(UTC)` call (network round-trip, embedding, BM25 indexing, graph write — easily seconds to minutes) is silently dropped on the next sync because the next `since` will be past their `updated` timestamp.

This is the same class of bug the PR claims to fix (silent data loss via a cursor that races the source-of-truth). Recommended pattern:

```python
# at the top of _run_connection_sync, BEFORE the fetch
fetch_started_at = datetime.now(UTC)
...
# in finally:
last_synced_at=(fetch_started_at if status in ("success", "partial") else None),
```

This shifts the boundary the right way: re-fetching one document on the next sync (cheap, dedup handles it via `content_hash`) is strictly better than losing a document forever.

### BLOCKER / WARNING — Confluence has the exact same minute-precision trap; the PR only patches Jira

`src/metatron/connectors/confluence.py:126`:
```python
cql += f' AND lastModified > "{since.strftime("%Y-%m-%d %H:%M")}"'
```

Same minute granularity, no post-filter — so the same "every sync re-emits boundary-minute docs" symptom you describe in the PR body still happens on Confluence. Either:
- lift the post-filter into a small helper (`filter_after_cursor(documents, since)`) and apply it in Confluence + Jira (and any future connector that uses minute-resolution CQL/JQL), or
- add a Confluence-side `_extract_updated_at` equivalent.

Either way, leaving Confluence unpatched here means the prod symptom ("every sync returns N fetched, same N docs each time") will continue for Confluence connections.

### WARNING — Post-filter runs *after* expensive `_issue_to_document` parse

`src/metatron/connectors/jira.py`:
```python
for raw_issue in issues:
    try:
        doc = self._issue_to_document(raw_issue, workspace_id)   # expensive (ADF + changelog + comments)
        if since is not None:
            updated_at = self._extract_updated_at(raw_issue)
            if updated_at is not None and updated_at <= since:
                continue                                          # we just threw away that work
        documents.append(doc)
```

`_extract_updated_at` is intentionally cheap (raw dict access); `_issue_to_document` is not. Flip the order:
```python
for raw_issue in issues:
    if since is not None:
        updated_at = self._extract_updated_at(raw_issue)
        if updated_at is not None and updated_at <= since:
            continue
    try:
        doc = self._issue_to_document(raw_issue, workspace_id)
        documents.append(doc)
    except Exception as e:
        logger.warning("jira.issue.error", error=str(e))
```

Matters most on the exact scenario the PR is fixing — incremental syncs where the bulk of `issues` in the JQL window get dropped by the post-filter.

### WARNING — Multi-tenant blast radius in `/admin/reindex`

`src/metatron/api/routes/admin.py`:
```python
await conn.execute(text("UPDATE connections SET last_synced_at = NULL"))
await conn.execute(text("UPDATE raw_documents SET qdrant_synced = false, ..."))
```

No `WHERE workspace_id = :ws`. A single admin call resets **every workspace's** cursor and re-ingestion state. The endpoint itself doesn't accept `workspace_id`, so this is by design — but the docstring should state explicitly that this is a **global** reset, and ideally the endpoint should require a `workspace_id` (or an explicit `all_workspaces=true` flag) so an operator running it in a multi-tenant deployment can't accidentally trigger a fleet-wide re-embed/re-graph storm. Right now `X-Confirm-Reindex: yes` is the only safety net and it doesn't convey scope.

### WARNING — `error_message=None` is unconditional; convention diverges from `last_synced_at`

`src/metatron/storage/postgres.py` (`update_connection_status`):
```python
set_parts = ["status = :status", "error_message = :error_message"]
...
if last_synced_at is not None:
    set_parts.append("last_synced_at = :last_synced_at")
```

You've introduced the "`None` means leave the column unchanged" semantics for `last_synced_at` (which is core to the B1 fix — failed syncs must not advance the cursor), but `error_message` still uses "`None` means write NULL". Result: the pre-flight `update_connection_status(connection_id, status="syncing")` call in `trigger_sync` silently NULLs out any prior error context the moment a retry begins, even if the new sync also fails. Either:
- apply the same conditional-SET pattern to `error_message`, or
- make the divergence explicit in the docstring and have callers pass `error_message=""` (or a sentinel) when they explicitly want to clear it.

### SUGGESTION — `/admin/reindex` builds a fresh `PostgresStore`

```python
settings = Settings()
store = PostgresStore(settings.postgres_dsn)
try:
    async with store._engine.begin() as conn:
        ...
finally:
    await store.close()
```

The app already keeps a pooled `PostgresStore` on `request.app.state` — same DI helper the connections route uses. Spinning up a fresh engine (with its own pool) for two `UPDATE`s is wasteful and reaches into `_engine` (a private attribute). Use the DI store; engine teardown disappears.

### SUGGESTION — `_run_connection_sync` finally-block comment is already stale

```python
# Cursor advancement happens via update_connection_status above
# (connections.last_synced_at = NOW). The legacy file-based
# SyncState write has been removed (MTRNIX-332) — it was unreliable
# under multi-replica / read-only-FS deployments and PG is now the
# single source of truth for the incremental cursor.
```

Once the follow-up cleanup PR drops `sync_state.py`, this block is dead weight. The "why" belongs in the commit message / ADR, not the source. Trim to one line ("Cursor advanced above via `update_connection_status`.") or remove entirely — the docstring already explains it.

### Test gaps

1. `test_run_connection_sync_failed_does_not_advance_cursor` patches the registry but leaves `mark_documents_synced_by_source`, `process_all_unsynced_graphs`, and `update_sync_log` unpatched. If any downstream collaborator is unavailable in CI, the test still passes (`status == "error"` is the outcome of any internal failure, not just the seeded `RuntimeError`). Tighten by also asserting `errors_list` / `update_sync_log` was called with the seeded error string, or patch the collaborators.

2. No test covers **the actual race the PR is meant to close**: cursor read from PG = X, fetch returns docs whose `updated > X`, sync succeeds, PG cursor advances. Currently you have "force_full bypasses cursor" and "null cursor → full fetch" but not "non-null cursor → since == that cursor". `test_run_connection_sync_default_uses_pg_cursor` is close but only asserts `since` is forwarded; it doesn't assert the cursor was *advanced* on success. Add an assertion that `update_connection_status` was called with `last_synced_at` ≈ now() after success.

3. No test for the `force_full=True` + failed-sync path. Per the docstring ("The cursor is still stamped on success"), force_full + failure must also NOT advance — worth pinning down.

### Nit

- `_extract_updated_at` catches `AttributeError` but the chained `.get("fields") or {}` short-circuit already neutralises non-dict shapes. Either narrow the parameter type to `Mapping[str, Any]` and drop the broad except, or document why you're hedging against malformed dicts coming out of `enhanced_jql`.
- The `_empty_ingest_result()` helper uses `MagicMock` instead of the real `IngestResult` dataclass — silent breakage if fields are renamed. Build the real dataclass if it's a dataclass.

---

Approach is right (PG as single source of truth + sub-minute post-filter); execution has two correctness gaps (NOW-as-cursor and Confluence-not-patched) that should land in the same PR before this ships to prod.

PR #120, opened 2026-05-18
feat(MTRNIX-336): RAG generation telemetry for fine-tuning bootstrap

pahomm:
Persist every LLM completion that flows through Metatron into a new PostgreSQL table (llm_generation_log) so a fine-tuning dataset can be assembled later for replacing the RAG answer-generation call with a local model. Bootstrap-only — no on-prem migration, no streaming capture, no retention worker.

Coverage:
- 11 of 12 LLM call sites instrumented via a single keyword-only call_site kwarg on chat_completion()/chat_completion_with_retry(): rag_answer, resolve_query, translate_query, hyde, query_expansion, query_classifier, routing, translation_to_english, translation_to_russian, ner_extraction, mcp_action_planner. agent_smalltalk wired for consistency (legacy/deprecated). freshness_decision deferred — bypasses the public wrapper, tracked in audit doc as follow-up.
- Entry-points push a TelemetryContext via ContextVar: REST (rest), OpenAI-compat (oai_compat), MCP tool wrapper (mcp), ingestion pipeline (ingestion) — both _write_graph and _write_graph_strict for decoupled graph extraction, benchmarker (benchmark), offline eval script (eval). Source tag lets the export utility exclude eval/benchmark traffic from the FT dataset.

Design notes:
- Writes are synchronous via psycopg2 inside the worker thread — asyncio.create_task() is unsafe from asyncio.to_thread workers (no running event loop). Single round-trip per LLM call; event loop is never blocked.
- One row per public chat_completion() invocation. Fallback recorded as metadata.fallback_used; retries naturally produce N failure rows
  + one success row from chat_completion_with_retry.
- Per-workspace opt-out via workspaces.llm_telemetry_opt_out (default FALSE) with 60s TTL cache and single-flight per-workspace lock. Early-gate inside chat_completion skips prompt-snapshot construction when opt-out or kill-switch is active — privacy posture is "we don't process this prompt", not "we don't store it". emit_log re-checks under its own lock, so flips mid-call are safe.
- Master kill-switch METATRON_LLM_TELEMETRY_ENABLED=true by default; emit_log never raises, write failures are logged at WARNING.
- Empty provider responses become synthetic failures with error_class='EmptyResponse' so the chk_outcome_coherent CHECK constraint stays satisfied and downstream filters are trivial.
- metadata JSONB carries fallback_used/zero_tokens common keys plus per-site extras (rag_answer subtype freeform|team_workflow_schema + lang + doc_ids; ner_extraction text_truncated + doc_label). add_extra_metadata() helper merges instead of overwriting so multiple call sites in the same scope coexist.

Schema (migration 022):
- llm_generation_log: id, created_at, call_site, source, workspace_id, user_id, agent_id, correlation_id, provider, model, request_messages, response_content, prompt/completion/total_tokens, latency_ms, success, error_class, error_message, metadata. Indexes on (workspace_id, created_at DESC), (call_site, created_at DESC), partial on correlation_id WHERE NOT NULL. CHECK constraint: (success AND response_content IS NOT NULL) OR (NOT success AND error_class IS NOT NULL).
- workspaces.llm_telemetry_opt_out BOOL NOT NULL DEFAULT FALSE.

Export utility: scripts/export_llm_dataset.py emits openai-chat-ft (OpenAI fine-tune Chat format), openai-completion-legacy, or messages-only JSONL. Streams over created_at cursor. Filters out benchmark/eval sources by default; --include-eval opts in. --no-include-zero-tokens drops rows where every token count is 0.

Audit deliverable: docs/llm_call_sites_audit.md tabulates all sites with verdicts (fine-tuning-grade vs telemetry-only) and notes the freshness_decision deferral path.

New env vars (core/config.py):
- METATRON_LLM_TELEMETRY_ENABLED=true (master kill-switch)
- METATRON_LLM_TELEMETRY_RETENTION_DAYS=0 (placeholder, no worker yet)
- METATRON_LLM_TELEMETRY_OPT_OUT_CACHE_TTL_SECONDS=60

Tests (31 new, 4 retry tests adjusted):
- test_llm_telemetry: set/reset/nested scopes, update_retrieved_context mutation, kill-switch and opt-out skip, store-failure swallow, sync emit from asyncio.to_thread (B1 regression guard), zero_tokens flag.
- test_chat_completion_telemetry: success/empty-content/fallback/retry/ no-fallback/empty-usage paths, fallback_used metadata.
- test_export_dataset: each format shape, success-only and benchmark/eval/zero-tokens filter defaults.

Verified end-to-end on dev stand: REST /api/v1/chat, /v1/chat/completions, MCP metatron_search via streamable-http, Jira+Confluence ingestion all produce correctly-tagged rows with shared correlation_id per request.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -250,6 +250,19 @@ METATRON_FRESHNESS_KB_STALE_AFTER_DAYS=90           # higher than memory's 30 (K
 
 
 # =============================================================================
+# LLM generation telemetry — fine-tuning bootstrap (MTRNIX-336)
+# Persists every chat_completion() into the llm_generation_log table so the
+# RAG answer-generation call can later be replaced with a local fine-tuned
+# model. Privacy note: by default every workspace contributes rows; set
+# workspaces.llm_telemetry_opt_out=true per workspace for opt-out, OR flip
+# the master kill-switch below.
+# =============================================================================
+METATRON_LLM_TELEMETRY_ENABLED=true                 # master kill-switch; false → emit_log is a no-op end to end
+METATRON_LLM_TELEMETRY_RETENTION_DAYS=0             # placeholder for future cleanup worker (0 = infinite)
+METATRON_LLM_TELEMETRY_OPT_OUT_CACHE_TTL_SECONDS=60 # TTL for the per-workspace opt-out flag cache
+
+
+# =============================================================================
 # Misc / Ops
 # =============================================================================
 ALLOW_CLEANUP=false                   # gate destructive /api/v1/admin/cleanup* endpoints
--- a/docs/llm_call_sites_audit.md
+++ b/docs/llm_call_sites_audit.md
@@ -5,6 +5,14 @@
 
 Every LLM call site in Metatron Core, with telemetry verdict and wiring status.
 
+## Backward-compatible default for external callers
+
+`chat_completion(..., call_site: str = "unknown")` — keyword-only, **defaults
+to "unknown"**. Plugins / enterprise extensions that import the public
+wrapper without updating their call sites still work; their rows just land
+under the `unknown` bucket and can be filtered or relabelled later. In-tree
+callers always pass a stable label from the table below.
+
 | # | `call_site` label | File | Verdict | Wired | Notes |
 |---|---|---|---|---|---|
 | 1 | `rag_answer` | `retrieval/search.py` | **fine-tuning-grade** | ✓ | Primary target. Two branches: `use_schema=True` (`subtype="team_workflow_schema"`) and regular free-form (`subtype="freeform"`). `update_retrieved_context` called before both LLM invocations. |
@@ -48,9 +56,33 @@ Every LLM call site in Metatron Core, with telemetry verdict and wiring status.
 Formats: `openai-chat-ft` (default), `openai-completion-legacy`, `messages-only`.
 Benchmark/eval rows excluded by default (`--include-eval` to opt in).
 
+## Telemetry safety / hardening behaviours
+
+- **Per-message content cap** — `emit_log` truncates each request message
+  `content` and the response `content` to 8 000 chars (matches the NER
+  upstream cap in `extract_graph_from_text`). Truncation is flagged in
+  `metadata.message_truncated` / `metadata.response_truncated`. Prevents
+  multi-GB JSONB growth on the high-frequency NER path.
+- **Lazy prompt snapshot** — the `messages` argument to `emit_log` can be a
+  zero-arg callable; `emit_log` materialises it only after the opt-out
+  re-check. So when a workspace flips `llm_telemetry_opt_out=true`
+  mid-call, the prompt copy never leaves the upstream `Message` list.
+- **Bounded per-workspace lock cache** — opt-out single-flight uses an
+  `OrderedDict` capped at 1 024 entries (LRU eviction). Locks for evicted
+  workspaces are GC'd when their last holder releases them.
+
 ## Open follow-ups
 
-- `freshness_decision` instrumentation — `LLMBackedDecisionEngine.decide` bypasses the public wrapper.
-- Background writer thread for `emit_log` if PG round-trip overhead is unacceptable on NER path.
-- Retention/cleanup worker for `llm_generation_log` (env var `METATRON_LLM_TELEMETRY_RETENTION_DAYS` is reserved).
+- `freshness_decision` instrumentation — `LLMBackedDecisionEngine.decide`
+  bypasses the public wrapper. **TODO: file Jira issue and link here.**
+  Until then, this audit doc is the canonical placeholder. The fix is
+  either (a) route the engine through `metatron.llm.chat_completion(...,
+  call_site="freshness_decision", provider=<configured>)`, moving the
+  freshness-specific provider override into the wrapper's resolution, or
+  (b) add a parallel provider-level telemetry hook callable from outside
+  the public wrapper.
+- Background writer thread for `emit_log` if PG round-trip overhead is
+  unacceptable on the NER path.
+- Retention/cleanup worker for `llm_generation_log` (env var
+  `METATRON_LLM_TELEMETRY_RETENTION_DAYS` is reserved).
 - Workspace-settings UI/API for `llm_telemetry_opt_out`.
--- a/docs/llm_call_sites_audit.md
+++ b/docs/llm_call_sites_audit.md
@@ -1,0 +1,56 @@
+# LLM Call Sites Audit — MTRNIX-336
+
+**Date:** 2026-05-15
+**Status:** Implemented (11 of 12 call sites wired; `freshness_decision` deferred)
+
+Every LLM call site in Metatron Core, with telemetry verdict and wiring status.
+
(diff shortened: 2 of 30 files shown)


godedok:
Reviewed against the MTRNIX-336 DoD and `CLAUDE.md` constraints (layer rules, plugin back-compat, workspace isolation, async-everywhere). The bootstrap is comprehensive and the test coverage is solid (the `asyncio.to_thread` regression guard in `test_llm_telemetry.py` is the right thing to assert). A few items I'd want addressed before this rolls to a tenant:

### Blocking

1. **Public API break in `metatron.llm.chat_completion`.** `call_site` is now a *required* keyword-only argument (`src/metatron/llm/__init__.py:108`). The repo-level rule in `CLAUDE.md` is `Do NOT … Break backward compatibility — no plugins = core works as before`. Any plugin (or enterprise extension) that today imports `chat_completion` and doesn't pass `call_site=...` will raise `TypeError` at call time. Two acceptable resolutions:
   - Default `call_site: str = \"unknown\"` (writes a row with `call_site='unknown'` — visible in the dataset, no crash for external callers), or
   - Keep it required but coordinate a version bump + release note + plugin-author migration guide.
   I'd prefer option 1: the audit doc already lists `agent_smalltalk` as `skip / deprecated` but we still tagged it — same idea works for unknown external callers.

2. **`request_messages` JSONB has no size cap.** `src/metatron/storage/llm_generation_log.py` writes whatever the call site hands over. For `ner_extraction` (per-document, post-truncation 8 000 chars → maybe 30–40 KB JSON), the table will grow by ~30–40 KB × every chunk × every ingestion run. Over a month of normal Confluence/Jira sync this is multi-GB of data the retention worker (still a placeholder per `METATRON_LLM_TELEMETRY_RETENTION_DAYS=0`) will not reclaim. Two options:
   - Cap message content to N chars (8 000 is reasonable, matches the NER truncation), and stamp a `message_truncated: true` flag in metadata when truncation happened.
   - Or land a basic retention worker in this PR (the env var is already there).
   If neither lands, please add a `TODO` in `migrations/versions/022_llm_generation_log.py` and an operator note in the audit doc — right now this can fill PG silently.

### Substantive

3. **`emit_log` is duplicated 5× inside `chat_completion`.** `src/metatron/llm/__init__.py:160-280`: success / empty-content / no-fallback / fallback-success / fallback-empty / fallback-failure each construct the same kwargs by hand. This is a maintenance trap — the next person who adds a field (e.g. `tenant_id` for multi-tenant ops) will inevitably miss a branch and the export will be subtly wrong for the missed path. A tiny `_emit(...)` helper bound to `(call_site, messages_for_log, t0)` would collapse all five sites to one-liners.

4. **Unbounded `_opt_out_per_ws_locks` growth.** `src/metatron/llm/telemetry.py:106`: `threading.Lock` objects are added to `_opt_out_per_ws_locks` on first miss for each workspace and never evicted. For ASOC/pilot scale this is negligible, but if Metatron ever fronts hundreds of thousands of short-lived workspaces (test fixtures, ephemeral tenants) this dict leaks. Consider a simple LRU cap or `weakref.WeakValueDictionary` of locks.

5. **Race window leaves prompt in memory.** The `is_telemetry_writable()` gate at `__init__.py:155` builds `messages_for_log` when telemetry is on at call entry; if a workspace flips `llm_telemetry_opt_out=true` *during* the LLM call, `emit_log` re-checks under the lock and drops the row — good. But the prompt copy is already in process memory for the duration of the call. The privacy story says \"we don't process this prompt when opted out,\" which is true at start-of-call, but a mid-call flip means we briefly do hold a copy. Doc this race in the audit (it's a small honest caveat, not a bug), or invert the logic so messages_for_log is built lazily right before `emit_log`. Lazy build is the cleaner fix and changes the gate from optimistic to correctness.

6. **MCP wrapper manually enters/exits the context manager.** `src/metatron/mcp/server.py:174-180`: `telemetry_cm.__enter__()` / `telemetry_cm.__exit__(None, None, None)` in the `finally`. When `handler(...)` raises, `__exit__` is called with `(None, None, None)` — the contextmanager generator's `try/finally` still runs the `current_telemetry_ctx.reset(token)`, so this is functionally correct, but it's brittle: if anyone ever adds exception-aware cleanup inside `set_telemetry_context`, this site silently won't get it. Prefer wrapping the body in an actual `async with` (or `with` since it's a sync context manager — async wrapper around a sync CM is fine via `contextlib.AbstractContextManager`). Same applies to the `agent_id` token reset block right below — both could share one cleaner exit.

7. **`_write_graph_strict` was split into `_write_graph_strict` + `_write_graph_strict_inner`.** `src/metatron/ingestion/pipeline.py:755-770`. Functionally fine, but the split exists *only* to avoid re-indenting the body inside a `with` block — the inner function's docstring says exactly that. This is the kind of artificial helper that immediately rots: next person to touch the body will wonder why it's split, miss the comment, and start adding logic to the wrong function. Either: (a) eat the indent and inline, (b) name it `_write_graph_strict_unwrapped` so the reason is in the name. I'd just inline.

### Minor

8. `src/metatron/mcp/server.py:242-244`: the `_wrap_tool_with_activity` call was reflowed from 3 lines to 1 in this PR. It looks like a formatter side-effect — unrelated to telemetry. Not blocking, but mixing it in makes the diff harder to bisect later.

9. `migrations/versions/022_llm_generation_log.py:160`: `llm_telemetry_opt_out` is added with `nullable=False, server_default=text(\"false\")` — good for new installs and existing rows. But the migration is not reversible on data: `downgrade()` drops the column with no archive step. Since the column is just a flag with operator intent, this is fine, but worth noting in the migration docstring.

10. Audit doc `docs/llm_call_sites_audit.md` says `freshness_decision` is deferred because it bypasses the public wrapper. That's the right call. Could you file the follow-up ticket as a real Jira issue and link it from the audit doc and from the new env-var description, so it doesn't get lost? Right now `LLMBackedDecisionEngine.decide` is the one production code path silently missing from the FT dataset, and it would be the first thing someone discovers two months from now when the dataset is short.

### Positives worth keeping

- The CHECK constraint `chk_outcome_coherent` on success/response_content/error_class is exactly the right invariant — downstream filtering becomes trivial.
- Cursor-based pagination in `scripts/export_llm_dataset.py` (no `OFFSET`) is the right choice for a growing table.
- The opt-out single-flight via per-workspace lock is a clean pattern; the `_is_opted_out` slow-path re-check under the lock is correct.
- The `extra_metadata` Python-attr → `metadata` DB-column mapping with the explanatory comment in `pg_models.py` is the right tradeoff against SQLAlchemy's reserved name.
- The empty-content → synthetic `EmptyResponse` failure row keeps the CHECK constraint satisfied without forcing callers to handle a new exception type — well done.

### Suggested next steps

- (Blocking) Decide on `call_site` default vs. required + migration note.
- (Blocking) Add a content-size cap or commit to the retention worker in this PR.
- (Substantive) Extract the `emit_log` invocation helper.
- (Substantive) Decide on the messages_for_log race and either doc or fix.
- (Substantive) Lazy-bound per-workspace lock dict.
- File the `freshness_decision` follow-up ticket.

Happy to look at the next iteration.

PR #126, opened 2026-05-26
feat(MTRNIX-341): add /finops/active-users endpoint

pahomm:
Returns active_users (distinct authenticated principals who reached call_site='rag_answer' via oai_compat or rest in the window) and period_queries (total queries from the same filtered slice) so the dashboard can compute Avg Queries/User from a self-consistent slice rather than mixing this endpoint with /finops/time-savings (which includes MCP and other channels).

Reads existing llm_generation_log (MTRNIX-336) — no migration, no new env vars. Graceful degradation on DB errors and opt-out workspaces (returns 0/0). Matches existing FinOps auth posture (any authenticated user, no workspace_id JWT cross-check — known limitation tracked separately).

### DIFF
--- a/src/metatron/api/routes/.claude/finops.md
+++ b/src/metatron/api/routes/.claude/finops.md
@@ -59,3 +59,66 @@ Empty days in the period are filled with zeros.
 
 ## Future: Cost Savings (MTRNIX-169, separate task)
 Will add `GET /api/v1/finops/cost-savings` — compares OpenAI API cost vs Metatron infrastructure cost per document. Different formula, different endpoint, same FinOps page.
+
+---
+
+# FinOps — Active Users API (MTRNIX-341)
+
+## Overview
+Backs the "Unique Users" and "Avg Queries/User" dashboard cards. Counts distinct
+authenticated principals who completed a RAG answer in the window, plus total queries
+from the same filtered slice.
+
+## Endpoint
+`GET /api/v1/finops/active-users?workspace_id={id}&days={30}`
+- `days`: 1..365, default 30 (same contract as `/time-savings`).
+
+## Response Schema
+```json
+{
+  "period_days": 30,
+  "active_users": 42,
+  "period_queries": 1337
+}
+```
+- `active_users` — `COUNT(DISTINCT user_id)` over the filtered slice.
+- `period_queries` — `COUNT(*)` over the SAME slice; frontend computes
+  `Avg Queries/User = period_queries / active_users`. Do NOT divide
+  `/time-savings.total_queries` (different slice — includes MCP/other) by `active_users`.
+
+## Data Source
+- Table: `llm_generation_log` (PostgreSQL, migration 022, MTRNIX-336) — NOT `query_traces`.
+- Filter: `source IN ('oai_compat', 'rest')`, `call_site = 'rag_answer'`, `user_id IS NOT NULL`,
+  `created_at >= now() - days`.
+- `user_id` here comes from the authenticated `TelemetryContext` (trustworthy), unlike
+  `query_traces.trace->>'user_id'` which defaults to the literal `"user"` for MCP calls.
+
+## Why these filters
+- `source IN ('oai_compat','rest')` — excludes MCP (no per-user identity), ingestion,
+  freshness, benchmark (system traffic, no user_id).
+- `call_site='rag_answer'` — only the synthesis LLM call counts as "used RAG"; users whose
+  pipeline failed earlier (resolve/translate/expand/classify) are not counted.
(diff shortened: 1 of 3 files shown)


godedok:
## Review

**Verdict:** Approve with one discussion point and a few nits. SQL, async pattern, error handling, and tests are solid. The interesting question is *where* this endpoint should live.

---

### Discussion — `routes/finops.py` is on the legacy chopping block

`docs/LEGACY.md §5` lists `api/routes/finops.py` as **"EXTRACT or REMOVE — if Control Center wants usage metrics, it can pull from `query_traces` directly."** Root `CLAUDE.md` Do-NOT list says verbatim: *"Build new features in `channels/`, `finops`, `api/routes/chat.py` — these are legacy."*

This PR adds a new endpoint, new response model, in that module. Two options:

- **(preferred) Move to `routes/dashboard/overview.py`.** It already hosts the dashboard-aggregation pattern (`/dashboard/overview`, `/dashboard/activity`) and has `get_valid_workspace()` for proper workspace gating. Same SQL, different mount; `/dashboard/active-users` is also a more honest URL given the Jira DoD literally says "dashboard tab".
- **Accept the legacy-extend cost** with an explicit note in the PR description and an issue tracking the eventual move — defensible only if the frontend already hardcodes `/api/v1/finops/active-users` and changing both sides now is too costly for the pilot timeline.

Either way, the choice should be conscious. Right now the PR silently extends a module that's flagged for removal.

---

### Warnings

**W1 — Workspace not validated against JWT.** `workspace_id` is a `Query(...)` param; any authenticated user can read any workspace's stats. PR body acknowledges this matches the existing FinOps posture. OK that the PR doesn't fix it, but if it moves to `/dashboard/*` (see above) the fix is free via `get_valid_workspace()`.

**W2 — "Opt-out returns 0/0" is implicit, not enforced in the SELECT.** The docstring claims opt-out workspaces return 0/0, but the SQL has no `JOIN workspaces ON llm_telemetry_opt_out=false`. It works only because `llm/telemetry.py` skips inserts at write time. A workspace that toggles opt-out *after* accumulating rows will still count those legacy rows. Either add a one-line docstring note ("opt-out enforced upstream at write time; pre-opt-out rows remain counted") or JOIN for defense-in-depth. The docstring note is cheap and accurate.

**W3 — `call_site='rag_answer'` is a magic string** repeated in helper, docstring, and SQL-snapshot test. If MTRNIX-336 ever renames it, three places break. A module-level constant (or import from `metatron.llm.telemetry`) is more defensive.

---

### Suggestions

**S1 — Filter-snapshot test is fragile** (`test_query_filters_are_correct`). It greps a lowercased compiled SQL string for substrings; a regression where `source IN ('oai_compat',)` (rest dropped) would still pass because `"source in"` is still there. Tighten to `"in ('oai_compat', 'rest')"` or inspect `stmt.whereclause` directly.

**S2 — `tuple[int, int]` is positional and unlabeled at call sites.** A `NamedTuple` / small dataclass (`ActiveUsersCounts(active_users, period_queries)`) prevents silent swap. Tiny change, real safety.

**S3 — Tests bypass `OptionalAuthMiddleware`** — `TestActiveUsersHttpRoute` mounts `router` in a bare `FastAPI()` instead of going through `create_app(settings)`. Auth regressions on this endpoint wouldn't be caught. Acceptable for unit-scope but worth one integration smoke test if/when this moves.

**S4 — `routes/.claude/finops.md` should document the new endpoint** (the existing `/time-savings` is there). Presumably pending the `documenter` phase — just flagging so it isn't missed at merge.

---

### Verified — non-issues

- ORM model fields (`workspace_id`, `user_id`, `source`, `call_site`, `created_at`) all exist on `LLMGenerationLogRow` per `pg_models.py:335` and migration 022. ✓
- Index `ix_llm_log_ws_created (workspace_id, created_at DESC)` covers the range scan; remaining filters are post-scan. Adequate for dashboard cadence. ✓
- Async pattern (sync ORM + `asyncio.to_thread`) matches `_fetch_time_savings`. ✓
- Graceful DB-error degradation matches `_fetch_time_savings`. ✓
- `pytest asyncio_mode=auto` — async tests don't need a marker. ✓
- The `.github/workflows/Deploy DEV.yml` / `install/docker-compose.yml` deletions in the diff are **branch drift** (develop is ahead via `ae10e64` / `7dc44fe`), not this PR's work — a rebase before merge will clean them up.

---

### Recommendation

Before merge: resolve the discussion point (move vs. accept-with-issue) and W2/W3 (~5 min total). S1–S4 are nice-to-have follow-ups. Implementation is otherwise ready.

PR #127, opened 2026-05-27
feat: access-checked ?workspace_id override for REST family B

pahomm:
Control Center pages (Agent Registry, Memory Inspector, Knowledge) did not react to workspace switching because agents/memory/knowledge endpoints resolved workspace strictly from the JWT (and fell back to the default workspace for admin tokens). Introduces resolve_workspace_id, a query-aware resolver with a JWT access check, and attaches it via a router-level workspace_scope dependency on the agents/memory/knowledge/snapshots routers so ?workspace_id is declared in OpenAPI for every route (incl. resource-keyed ones — variant B) and 403 is enforced uniformly. Empty workspace_ids is treated as unrestricted to match get_workspace_id and the AUTH_ENABLED=false case (default admin token has []). MCP path and build_telemetry_context_cm are intentionally unchanged.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -409,7 +409,7 @@ Upstream: metatron on port 8000, healthcheck at /health
 ## Conventions
 - Async everywhere: `async def` for handlers, DB calls, LLM calls
 - Config via pydantic-settings, env vars with METATRON_ prefix
-- Workspace isolation: all queries filtered by workspace_id (JWT claim)
+- Workspace isolation: all queries filtered by workspace_id (JWT claim by default). REST family-B endpoints (agents/memory/knowledge) also accept an optional, JWT-access-checked `?workspace_id` query param overriding the auth-derived default (`resolve_workspace_id` in `api/dependencies.py`); a caller may only target a workspace its token grants (`*` or membership), else 403. MCP stays strictly auth-derived.
 - Graceful degradation: `_safe_call()` — Neo4j down → search works without graph
 - Factory pattern: `create_app(settings)` for isolated testing
 - Logging: structlog
--- a/src/metatron/api/.claude/claude.md
+++ b/src/metatron/api/.claude/claude.md
@@ -171,7 +171,7 @@ Memory REST API endpoints (workspace-scoped, RBAC-gated):
 
 **DI helper:** `get_memory_service(request)` — per-workspace cache on `app.state.memory_services`; shared PG engine on `app.state.memory_pg_engine`. Now also wires `freshness_store` (required for the review endpoints) and passes `pg_store` to `MemorySearchService` for graph-leg post-filter parity with the MCP path (MTRNIX-324).
 
-**Workspace resolution:** uses `get_workspace_id(request)` exported from `api.dependencies` (workspace always derived from auth, never from body/query).
+**Workspace resolution:** uses `resolve_workspace_id(request)` from `api.dependencies` — auth-derived by default, with an optional access-checked `?workspace_id` query override on REST family-B endpoints (agents/memory/knowledge); a caller may only target a workspace its JWT grants (`*` or membership), else 403. `get_workspace_id` remains the auth-only base (still used by telemetry / `build_telemetry_context_cm`).
 
 **Schemas:** pydantic v2 request/response models live inline in `routes/memory.py` per codebase convention.
 
--- a/src/metatron/api/dependencies.py
+++ b/src/metatron/api/dependencies.py
@@ -9,7 +9,11 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 from uuid import uuid4
 
-from fastapi import Request  # noqa: TC002 — FastAPI Depends parameters need runtime type
+from fastapi import (  # noqa: TC002 — FastAPI Depends parameters need runtime type
+    HTTPException,
+    Query,
+    Request,
+)
 
 from metatron.core.config import Settings  # noqa: TC001 — runtime annotations in function bodies
 from metatron.llm.telemetry import set_telemetry_context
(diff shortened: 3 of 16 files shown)


godedok:
## Code review — `feature/workspace-scoping-control-center`

Effort: extra-high, recall-mode (catch every plausible bug). 15 findings, ranked by severity.

---

### 🛑 CRITICAL — security

**1. Empty `workspace_ids: []` is fail-open** — `src/metatron/api/dependencies.py:76`

```python
if not allowed or "*" in allowed or requested in allowed:
    return str(requested)
```

The `not allowed` branch grants access to **any** requested workspace when the caller's JWT has `workspace_ids=[]`. The docstring justifies this as "matching `get_workspace_id`'s fallback semantics", but the two are NOT equivalent — `get_workspace_id` confines such callers to a **single** `default_workspace_id`, whereas `resolve_workspace_id` accepts an arbitrary id from the query string.

Reachable via the real login path: `UserStore.create_user(..., workspace_ids=None)` produces a user with no `user_workspaces` rows. `auth/routes/auth.py:52` then issues a JWT with `workspace_ids=[]`. After this PR, that user can call `GET /api/v1/memory/records?workspace_id=victim-ws` and read another tenant's memory. Tests `test_empty_workspace_ids_is_unrestricted` and `test_no_user_state_is_unrestricted` encode this as the intended contract — making the eventual fix break those tests.

Recommended: require non-empty `workspace_ids` and `"*" in allowed or requested in allowed`. Reject `[]` as 403 (or normalize at JWT verification time so the resolver never sees it).

**2. Path traversal via `workspace_id`** — `src/metatron/memory/snapshot.py:243`

```python
def _agent_dir(self, agent_id: str) -> Path:
    return self._snapshot_dir / self._workspace_id / agent_id
```

`resolve_workspace_id` returns `str(requested)` unmodified. A caller with `workspace_ids=["*"]` or `[]` calling `POST /api/v1/agents/<id>/snapshots?workspace_id=../../tmp` constructs `MemorySnapshotService(workspace_id="../../tmp", snapshot_dir="/data/snapshots")`. The path join produces `/data/snapshots/../../tmp/<agent_id>/<snapshot>.jsonl.gz` — a write outside the configured snapshot root. `Path` does not normalize automatically.

Recommended: validate `workspace_id` against a strict charset regex (e.g. `^[a-zA-Z0-9_-]{1,64}$`) before returning from `resolve_workspace_id`. This also closes finding #4.

**3. `AUTH_ENABLED=false` widens from "default-only" to "any workspace"** — `src/metatron/api/dependencies.py:74-78`

When auth is disabled, `OptionalAuthMiddleware` sets `request.state.user = {}`. Pre-PR, callers were confined to `default_workspace_id`; post-PR, they can target ANY workspace via `?workspace_id=`. This is the default deployment mode (`AUTH_ENABLED=false`).

---

### ⚠️ HIGH — security / correctness

**4. No input validation on the workspace_id string** — `src/metatron/api/dependencies.py:77`

`str(requested)` is returned raw. Downstream: Qdrant collection names (HTTP path component), filesystem paths (#2), Neo4j queries. Length and charset checks are missing. (`str()` on a `str` is also a no-op — minor cleanup.)

**5. `requested in allowed` is a substring match if `allowed` is a string** — `src/metatron/api/dependencies.py:76`

If a plugin auth backend ever returns `request.state.user["workspace_ids"]="ws-acme"` (string, not list — plausible schema drift for a single-workspace tenant), then `requested in allowed` becomes a substring check: `"ws" in "ws-acme"` is True. Add `isinstance(allowed, list)` defense before the membership test.

**6. Service caches grow without bound** — `src/metatron/api/dependencies.py:133-201` (and the 4 sibling factories)

Five `app.state.*_services` dicts (`memory_services`, `agent_registry_services`, `memory_health_services`, `memory_snapshot_services`, `raw_document_services`) have no eviction. A `*` admin (or empty-list caller, per #1) can pin unbounded entries — each carries a `MemoryQdrantStore` Qdrant client. Pre-PR the natural bound was the small set of workspaces derivable from auth; that bound is now removed and `resolve_workspace_id` does not validate workspace existence ("nonexistent workspace yields an empty result set downstream").

**7. Mutation endpoints also accept `?workspace_id`** — `src/metatron/api/routes/agents.py:68`

`Depends(workspace_scope)` is at the router level on `agents`, so `/start`, `/stop`, `/pause`, `/restore`, `/reset` honor `?workspace_id` for `*` admins. The PR description scopes the change to read/list ("did not react to workspace switching"); cross-tenant lifecycle is a sharper power and has no PR test coverage.

---

### 🔶 MEDIUM

**8. Three independent resolutions per request; resolved value discarded** — `src/metatron/api/dependencies.py:81-105`

`workspace_scope` resolves, the service factory (`get_memory_service` etc.) resolves again, and many handlers resolve a third time. The router-level dep returns `str` but the value is dropped. If a plugin middleware ever mutates the query string between dep layers, the access check and the data access can disagree (TOCTOU). Cleaner: stash on `request.state.workspace_id` once, then read everywhere.

**9. Tests enshrine the fail-open as a contract** — `tests/unit/api/test_resolve_workspace_id.py:62, 73`

`test_empty_workspace_ids_is_unrestricted` and `test_no_user_state_is_unrestricted` assert that empty `workspace_ids` grants the requested workspace. Once #1 is fixed these tests will fail — a future contributor seeing them may be tempted to keep the unsafe semantics. At minimum mark `xfail` with a tech-debt ticket.

**10. `?workspace_id=*` silently downgrades to the auth default** — `src/metatron/api/dependencies.py:72`

`requested == "*"` early-returns into `get_workspace_id`. A frontend passing `*` thinking "all my workspaces" (mirroring the JWT wildcard) gets a single workspace silently, no error. Result depends on caller's token: admin (`["*"]`) → `default_workspace_id`; non-admin (`["ws-a","ws-b"]`) → `"ws-a"`. Better to 400 explicitly.

**11. Whitespace `workspace_id` flows through** — `src/metatron/api/dependencies.py:72`

`?workspace_id=%20%20` → `requested = "  "` is truthy and != `"*"`, so it passes the early-return and is returned as the workspace id. Add `.strip()` and treat empty-after-strip as absent.

**12. Telemetry still uses `get_workspace_id`** — `src/metatron/api/dependencies.py:99`

`build_telemetry_context_cm` is unchanged by design. Today it's only used by `chat.py` (legacy, not family-B), so isolated. But the divergence means audit logs ≠ actual workspace accessed if anyone later wires telemetry into a family-B route. Worth a code comment on `build_telemetry_context_cm` warning future readers.

---

### 🟢 LOW

**13. Stale isolation claim in snapshots module docstring** — `src/metatron/api/routes/snapshots.py:5-9`

The leading docstring was updated but the sentence about cross-workspace isolation now relies on the new access check, not on the JWT being the only source of truth. Worth tightening the wording.

**14. `_resolve_workspace_id = get_workspace_id` alias is dead code** — `src/metatron/api/dependencies.py:81`

After this PR no src/ caller remains. Keeping a name that *starts with `resolve`* but means *strict auth-derived* invites future contributors to import the wrong one. Either delete or rename to e.g. `_auth_only_workspace_id`.

**15. Per-router opt-in fragility (altitude)** — `src/metatron/api/dependencies.py:105`

Each family-B router must remember to add `dependencies=[Depends(workspace_scope)]`. A new router (or a route mistakenly added elsewhere) silently uses `get_workspace_id` and drops the query override. The OpenAPI shape test is per-router. Consider stamping `request.state.active_workspace_id` once in an app-level middleware/helper so resolution is single-sourced for both REST and any future transport — and so MCP/OAI surfaces inherit the same access primitive when Permission Model v2 lands.

PR #130, opened 2026-06-02
feat(MTRNIX-372): OpenAI-compatible proxy LLM with context assembly

godedok:
## What

Adds an OpenAI-API-compatible **proxy LLM** surface so external agents (Hermes, OpenWebUI, custom OAI clients) gain memory + KB context without touching their system prompt, and every call is observable.

- **New endpoint** `POST /v1/proxy/chat/completions` — requires `X-Agent-Id`, static-key auth, SSE streaming.
- **`ProxyService.dispatch(mode=proxy|rag)`** — single orchestration entrypoint (D-020 / Confluence Q9, **A-full**): resolve agent → assemble context → inject into system message → forward to the agent's upstream LLM → tee usage → emit events.
- **Legacy `/v1/chat/completions`** refactored to delegate through `ProxyService.dispatch(mode=rag)` — RAG answer behaviour preserved (validations, workspace-existence 404, `user_id` attribution, plugin pipeline hooks all retained); golden-file SSE test as the gate.
- **`AgentContextAssembler`** extended: 4 sections (`constitution` reserved-empty / `preferences` / `relevant_memories` / `relevant_knowledge`), rule+optional-SLM query rewrite, per-stage timeouts, degradation tracking, raw `sections` dict.
- **`UpstreamLLMClient`** — httpx streaming, verbatim forwarding, error pass-through (never wraps upstream errors).
- **Tool-result enrichment** (Q7): `WorkspaceEntityTrie` + `ToolResultEnricher` — entity match → agent-scoped `get_memories_about_entity` → additive append to `<relevant_memories>` (preferences/knowledge untouched), 500ms hard cap, fail-open.
- **Observability:** 11 `proxy.*` activity event types + `PROXY_CALL_COMPLETED` EventBus event; `correlation_id` column on `agent_activity_log` for per-call tracing.
- **Schema (migration 023):** `agents.is_system` + per-workspace `system-chat` backfill; `llm_upstream_credentials` (Fernet-encrypted); `agent_activity_log.correlation_id`.

## Key design decisions (full log in spec §0)

- **A-full** single dispatch entrypoint; legacy SSE preserved bit-for-bit under golden test (A-lite was the regression fallback only).
- Identity = `X-Agent-Id` only; `model: metatron-proxy-{agent_id}` channel parked (OWUI cutover follow-up).
- Upstream keys: env `METATRON_PROXY_DEFAULT_UPSTREAM_KEY` + programmatic Fernet store; REST key mgmt deferred.
- `<constitution>` reserved-empty (no source wired this task).

Docs: design `docs/superpowers/specs/2026-05-28-mtrnix-372-proxy-llm-design.md` (read §0 first); plan `docs/superpowers/plans/2026-06-01-mtrnix-372-proxy-llm*.md`; glossary `CONTEXT.md`.

## Verification status

- ✅ **79 proxy/assembler unit tests pass**; ruff clean on all new/changed files.
- ⚠️ **Integration tests not yet run** (golden SSE, proxy e2e streaming, tool-result e2e, `correlation_id` query, agent-scoped graph) — require live PG/Qdrant/Neo4j (`make docker-up && make migrate`). To be executed post-review before merge.
- Note: repo-wide `make lint`/`make typecheck` baseline is pre-existing dirty (unrelated to this branch); `make test` aborts locally without optional test deps `respx`/`benchmark_qed`.

## Out of scope / parked

OWUI→proxy cutover + legacy retire, REST upstream-key management, `model`-field identity channel, enterprise `ProxyUsageMeter` rollup, memory extraction from tool results.


pahomm:
---
  🔴 BLOCKER

  1. Race on the shared UpstreamLLMClient.last_status — proxy/upstream.py:2137, proxy/service.py:1910
  UpstreamLLMClient is created once per app (app.py:231 → app.state.upstream_llm_client), and stream() writes self.last_status = response.status_code to an instance attribute. In _dispatch_proxy._generate() the status is read after the stream finishes:
  status = self._upstream.last_status or 0
  ok = error_reason is None and 200 <= status < 300
  With two concurrent requests, request A will read a last_status overwritten by request B. Result — a wrong upstream_status in proxy.upstream.completed/error and in PROXY_CALL_COMPLETED. This breaks exactly what correlation_id was introduced for (D8 —
  per-call tracing), and silently so. Tests don't catch the race (mock, single request).
  Fix: the status must be per-call — return it from stream() (e.g., carry it in ProxyStreamFrame or yield a leading "meta" frame), not store it on the singleton.

  2. Upward import + circular dependency L3 → L6 — proxy/service.py:1956
  async def _dispatch_rag(...):
      from metatron.api.routes.openai_compat import build_rag_stream
  proxy/ is declared as L3 (proxy/__init__.py: "— L3"), while api/routes/openai_compat is L6. On top of that, openai_compat.chat_completions itself builds ProxyService and calls dispatch (openai_compat.py:606), and ProxyService calls back into
  build_rag_stream → a full cycle masked by lazy imports. A direct violation of the CLAUDE.md rule "never import upward / Import from upper layers into lower".
  Fix: move build_rag_stream (the RAG-stream assembly logic) into a lower layer (L2 retrieval/ or a separate proxy/rag.py), and have both the L6 route and ProxyService depend down onto it.

  ---
  🟡 WARNING

  3. proxy_enabled defaults to True + un-run gate — config.py:701, PR body
  With the default True, after the merge on all existing deployments the legacy /v1/chat/completions will immediately start routing through ProxyService.dispatch(mode=rag). The byte-identity guarantee rests on the golden-SSE test, which, per the PR
  description, has not been run yet (needs live PG/Qdrant/Neo4j). Backward compatibility is a hard project rule. I recommend defaulting to False until the integration gate passes, or blocking the merge on running test_openai_compat_golden + test_proxy_e2e.

  4. mypy strict not verified — the PR only ran ruff on changed files. The new code has bare dict generics (payload: dict in app.py:368, -> list[dict] in _fetch_entity_memories), # type: ignore[attr-defined], Any. The project requires mypy strict — make
  typecheck needs to be run at least over the new modules.

  5. Silent swallowing of mid-stream upstream errors — proxy/service.py:1906
  Connection-level exceptions (an httpx error after headers are sent) are caught, logged, and the generator simply ends. The client gets a truncated stream with no error frame and no non-2xx signal (status 200 and headers already sent). This contradicts
  the claimed "error pass-through". At minimum, emit a terminal SSE error frame so the client can detect the failure.

  6. The is_system system-chat agent is created but unused — migrations/023:79
  The migration backfills one system-chat agent per workspace, include_system hides it — but no code path consumes it (rag mode runs with agent_id=None). The migration comment "backs the legacy no-X-Agent-Id fallback" does not match the code. Also
  gen_random_uuid() requires the pgcrypto extension. This is either dead data migration until a parked follow-up (OWUI/model channel), or the intent needs confirming.

  7. Object reconstruction on every request — app.py:262 _proxy_service_builder
  builder(workspace_id) is called on every proxy and every legacy chat request and each time constructs a fresh MemoryQdrantStore, MemorySearchService, AgentRegistryService, etc. (some are cached on app.state, but the Qdrant store is not). Potential
  connection churn/latency on the hot path. Worth caching the service/client by workspace_id.

  ---
  🟢 SUGGESTION

  - 8. Builder monkey-patch (app.py:408 svc._enricher_factory = ...): ProxyService.__init__ already accepts tool_result_enricher_factory — pass it via the constructor instead of mutating a private attribute. The comment in the code itself calls it a
  "monkey-patch".
  - 9. resolve_workspace_id(request) is called twice in proxy.py (in the route and in get_proxy_service) — resolve once.
  - 10. b'"tool_calls"' in frame.raw (service.py:1892) will false-positive if the substring appears in content text. Tolerable for an observability event, but worth noting.
  - 11. assert agent_id is not None (service.py:1791) as a runtime guard is stripped under python -O. An explicit check is better.
  - 12. inject_into_system only appends to the first system message; the multiple-system-messages case is uncovered.
  - 13. downgrade() does DELETE FROM agents WHERE is_system=true unconditionally — it would wipe any legitimately-created future system agents too.

  ---
  ✅ What's done well

  - Workspace isolation holds: resolve_workspace_id, Fernet keys with a cross-workspace isolation test, agent-scoping of get_memories_about_entity (prevents cross-agent bleed — a good catch).
  - Graceful degradation in the assembler with per-stage timeouts and degraded_sections; fail-open enrichment with a 500ms hard cap.
  - Pure functions inject_into_system / _render (no mutation, deepcopy).
  - Legacy-behaviour-preservation strategy with a golden gate; tidy feature-flag hygiene; strong unit coverage with proper mocking.
  - Alignment with D-020 (4 sections <constitution>/<preferences>/<relevant_memories>/<relevant_knowledge>).

  ---
  Verdict

  Do not merge until the 2 BLOCKERs are resolved (the last_status race, the upward import / L3→L6 cycle) and the integration gate is run (W3 — golden SSE + proxy e2e), without which the proxy_enabled=True default risks silently changing the legacy
  endpoint's behaviour on all deployments. W4 (mypy) is mandatory per project conventions.

  Want me to post this as inline comments on the PR (/code-review --comment), or go straight to checking the last_status race / circular import directly in the PR branch?

  A note: I created a temporary pr130.diff file in the repo root for the analysis — should I delete it?

PR #131, opened 2026-06-03
feat(rag-trace): full RAG debug trace — capture, persistence, read API

pahomm:
Capture a self-contained debug trace of every RAG request (OAI /v1/chat/completions and legacy /api/v1/chat, stream + non-stream) from user message to answer, persist it to PostgreSQL, embed its id in the answer, and expose read endpoints for a frontend viewer. Independent of llm_generation_log.

- retrieval/trace.py: RagTrace accumulator + pure phase builders + maybe_create_trace / append_trace_footer helpers; is_trace_enabled / footer_for.
- retrieval/search.py: rag_trace param + ~9 capture points (resolve/expand/translate/ classify/recall/merge_and_score/rerank/context_assembly/generation) + awaited persist; benchmarker return_trace path untouched.
- storage: RagDebugTraceRow ORM + migration 023 (rag_debug_traces, DESC index) + store/get/list_rag_traces_sync.
- api/routes/traces.py: GET /api/v1/traces/{id} + GET /api/v1/traces (viewer+, ws-scoped, reads NOT gated by the capture flag).
- config: METATRON_RAG_TRACE_ENABLED + METATRON_RAG_TRACE_FOOTER_ENABLED (both default true).
- candidate source derived from hit "type" (jira/confluence/...); generation phase records provider + resolved model name.
- tests: trace builders, search wiring, persist-on-llm-failure, endpoint, footer flag.
- docs/RAG_TRACE_FRONTEND.md: frontend interface + endpoint reference.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,19 @@
 ## [Unreleased]
 
 ### Added
+- feat(rag-trace): full RAG debug trace for answer debugging. Captures a self-contained per-request
+  trace (user message → resolve/expand/translate/classify → recall-per-channel with full candidate
+  text + scores → merge_and_score with per-signal breakdown → rerank → context_assembly with the
+  assembled prompt → generation with raw + final answer) and persists one `rag_debug_traces` row
+  (migration 024). New L3 module `retrieval/trace.py` (`RagTrace` accumulator + pure phase builders +
+  `maybe_create_trace`/`append_trace_footer` helpers); capture is threaded through
+  `hybrid_search_and_answer` via an optional `rag_trace` param (benchmarker `return_trace` path
+  untouched). New read API `GET /api/v1/traces/{trace_id}` + `GET /api/v1/traces` (viewer+,
+  workspace-scoped; reads NOT gated by the capture flag; 422 on non-UUID). The trace id is appended to
+  chat answers as a `— trace: <uuid>` footer. Rerank is captured after the ACL post-filter so
+  ACL-stripped chunks never enter the trace. New env vars `METATRON_RAG_TRACE_ENABLED` (default true)
+  and `METATRON_RAG_TRACE_FOOTER_ENABLED` (default true). Independent of `llm_generation_log`.
+  Frontend reference: `docs/RAG_TRACE_FRONTEND.md`. (No retention/cleanup yet — table grows unbounded.)
 - feat(memory-scopes): Phase 2 — session memory dual-write to PG + `lifetime` filter in Memory Inspector. `MemoryService.cache_session` now writes through to `memory_records` (Redis remains hot cache; PG write is best-effort, failures log `memory.session.pg_write_failed`). `GET /api/v1/knowledge/records` accepts `lifetime=persistent|session|all` (default `persistent`); response gains optional `session_id` and `ttl_expires_at` fields. Expired session rows are GC'd by a new `SessionGCPass` inside the existing freshness scheduled-scan loop (grace period via `METATRON_MEMORY_SESSION_GC_GRACE_HOURS`, default 24h). `/api/v1/memory/records?session_id=X` stays Redis-only for MCP back-compat.
 - feat(memory-scopes): unified knowledge view in Memory Inspector — new `GET /api/v1/knowledge/records?origin=agent|kb|all` endpoint that fans out across `memory_records` and `raw_documents`. `origin=all` uses `asyncio.gather` with partial-failure semantics (one leg down → 200 + `partial=true`; both down → 503). New L3 module `knowledge/` provides the `RawDocumentReadService` read facade. Zero schema migrations; zero new env vars; no changes to existing memory, MCP, or freshness surfaces. Audit: `docs/superpowers/2026-05-12-memory-scopes-audit.md`.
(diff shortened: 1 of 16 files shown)


godedok:
## Review

Solid feature overall: clean separation between pure phase builders (`trace.py`) and capture points in `search.py`, persist never raises into the answer path, partial trace is kept on LLM failure, benchmarker `return_trace` path untouched, reads are workspace-scoped with 404 on cross-workspace ids, and the list query matches the `(workspace_id, created_at DESC)` index. Test coverage of builders, wiring, and the endpoint is good. That said, there is one merge-stopper and a few real risks.

### BLOCKER

1. **Migration revision collision** — `migrations/versions/023_rag_debug_traces.py` declares `revision = "023"`, `down_revision = "022"`, but `develop` already has `023_add_proxy_support.py` with the exact same `revision = "023"` / `down_revision = "022"`. Alembic will fail on duplicate revision id and `make migrate` breaks for everyone. Rebase onto current `develop` and renumber to `024` with `down_revision = "023"`.

### WARNING

2. **ACL exposure via traces** — the rerank phase is captured right after `base = base[:k]`, i.e. **before** the ACL post-rerank defense-in-depth filter and `search_post_rerank` plugin hooks run. Chunks that those filters remove from the answer still land in the trace with full `text` and `kept: true`. On top of that, recall results are pre-filtered by the *requesting user's* groups, but the read endpoint only requires viewer + same workspace — so any viewer can read another user's trace, including content their own ACL groups would not return. In ACL-enabled workspaces this is a data-leak channel. Suggest: move the rerank capture after the ACL post-filter (or mark filtered candidates), and consider gating trace reads above viewer (or documenting the exposure explicitly as an accepted trade-off for the debug tool).

3. **Default-on capture with no retention** — `METATRON_RAG_TRACE_ENABLED` defaults to `true`, and each row stores full chunk texts for all recall candidates ×4 channels, the full merge pool, rerank pool, fragments, the assembled prompt, and both answers — easily hundreds of KB of JSONB per chat request. There is no TTL/cleanup, so the table grows unboundedly in production. A debug-capture feature should default OFF, or ship with a retention policy (scheduled deletion / per-workspace cap). The footer default-on also changes every user-visible answer on upgrade (clients parsing answers, golden/eval comparisons) — worth calling out as a behaviour change at minimum.

4. **`GET /api/v1/traces/{trace_id}` 500s on non-UUID input** — `trace_id` is a plain `str` path param, but the column is `UUID(as_uuid=False)`. A request like `/api/v1/traces/not-a-uuid` will hit PG with invalid uuid syntax and raise a DataError → 500 instead of 404/422. The endpoint test only passes because the store is mocked. Type the path param as `uuid.UUID` (or validate and 404 early).

5. **Traces with `workspace_id = NULL` are unreadable forever** — legacy `/api/v1/chat` passes `workspace_id=req.workspace_id` into the pipeline (pre-existing quirk), so when the client omits it the trace row may persist with `workspace_id = NULL`; the read queries filter `workspace_id == <resolved>` and NULL never matches. Either resolve the workspace before seeding the trace or skip persisting workspace-less traces — otherwise they're dead weight.

6. **Awaited persist adds write latency to the answer path** — `await _persist_rag_trace()` runs a synchronous PG write of a large payload (via `to_thread`) before the response returns, on every request while the flag is on. Consider fire-and-forget (`asyncio.create_task` with a done-callback for logging) — the trace is observability, the user shouldn't wait for it.

### SUGGESTION

7. **Inconsistent settings access** — `maybe_create_trace` / `is_trace_enabled` read `get_settings()` (cached singleton), while the persist gate in `search.py` reads the module-level `_s = Settings()` instance. Two different objects; they can diverge under runtime mutation/monkeypatching (your own `test_is_trace_enabled_reflects_setting` patches only the `get_settings()` instance). Use `get_settings()` in both places.

8. **`GET /{id}` omits `created_at`** — the frontend doc itself flags this as a quirk that will need a follow-up. Cheap to include row metadata (`created_at`, maybe `source`) alongside the JSONB now and save a round-trip change.

9. **In OAI streaming, `footer_enabled = is_trace_enabled() and ...` is redundant** — `append_trace_footer` already no-ops when `rag_trace is None` (which is exactly the capture-off case). Passing just the footer flag keeps one source of truth.

10. **`count` in the list response is page size, not total** — fine if intentional, but the name invites misuse for pagination math; either document it or rename (`returned`/`page_count`).

11. **`docs/RAG_TRACE_FRONTEND.md` is in Russian** — the rest of `docs/` is English; please translate for consistency.

12. **Docs not updated** — the two new env vars (`METATRON_RAG_TRACE_ENABLED`, `METATRON_RAG_TRACE_FOOTER_ENABLED`) belong in CLAUDE.md "Key Config", the new endpoints in the REST API section, and a CHANGELOG entry is missing.

13. **No test for `append_trace_footer` flag semantics** — `footer_for` is tested, but not the `enabled=False` / `trace=None` no-op branches that the routes actually rely on.

14. **Footer rendering edge case** — the footer line starts with an em-dash and immediately follows the sources block; frontends that split source lines on `" — "` may misparse it as a source entry. Worth a quick check in OpenWebUI before shipping with the footer on by default.

PR #137, opened 2026-06-09
feat(MTRNIX-396): KB autosync scheduler

godedok:
## MTRNIX-396 — Setup autosync functionality for KB

Implements a cron-driven scheduler so connections (Jira, Confluence, …) sync periodically without manual triggering.

Design grilled & agreed with the user; rationale captured in `docs/adr/2026-06-09-autosync-architecture.md` and glossary terms in `CONTEXT.md`.

### What it does
- New `connections.sync_cron` (cron schedule, `server_default '0 3 * * *'`) + `next_run_at` columns; `sync_logs.trigger` (`manual`|`scheduled`).
- In-process asyncio scheduler in the API lifespan (`api/autosync.py`), gated by `METATRON_AUTOSYNC_ENABLED` (default `true`).
- Per-tick: global due-SELECT → atomic conditional claim (multi-replica safe, also closes the pre-existing manual-path race) → bounded-concurrency spawn of the existing `_run_connection_sync`.
- Coalesce-to-one missed runs; cron interpreted in `METATRON_AUTOSYNC_TIMEZONE` (default UTC).
- `sync_cron` exposed on connection create/update/response with cron validation (422 on invalid).

### Defaults (intentional — see ADR)
- Autosync is **on by default**: new connectors and **backfilled existing connectors** get `0 3 * * *`. Channels (telegram/discord/slack) get `NULL`.

### Config
`METATRON_AUTOSYNC_ENABLED=true`, `AUTOSYNC_TIMEZONE=UTC`, `AUTOSYNC_POLL_SECONDS=60`, `AUTOSYNC_MAX_CONCURRENT=2`. New dep: `croniter`.

### Migration
Revision `025` (chains from `024`).

### Tests
19 new unit tests in `tests/unit/test_autosync.py`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 ### Added
+- feat: KB autosync — cron-driven periodic connection syncing, on by default (nightly 03:00) (MTRNIX-396)
 - feat(rag-trace): full RAG debug trace for answer debugging. Captures a self-contained per-request
   trace (user message → resolve/expand/translate/classify → recall-per-channel with full candidate
   text + scores → merge_and_score with per-signal breakdown → rerank → context_assembly with the
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -80,6 +80,7 @@ src/metatron/
 ├── app.py                    # Unified entry: asyncio.gather(API + bots)
 ├── api/
 │   ├── app.py                # create_app() factory, lifespan, plugin discovery
+│   ├── autosync.py            # AutosyncScheduler — in-process cron-driven connection sync loop (MTRNIX-396)
 │   ├── middleware.py          # OptionalAuthMiddleware (JWT gate)
 │   ├── dependencies.py        # FastAPI DI helpers
 │   └── routes/                # auth, chat, admin, skills, connections, documents,
@@ -233,6 +234,12 @@ Document flow: Connector → PostgreSQL (raw_documents) → Qdrant + Neo4j.
 PostgreSQL raw_documents table is the source of truth; Qdrant/Neo4j are derived stores.
 Graph extraction is decoupled from sync (process_all_unsynced_graphs, graph-process CLI).
 
+Migration 025 (`025_autosync_schedule.py`) adds `connections.sync_cron TEXT DEFAULT '0 3 * * *'`,
+`connections.next_run_at TIMESTAMPTZ NULL`, and `sync_logs.trigger TEXT NOT NULL DEFAULT 'manual'`.
+Connector rows are backfilled to nightly `0 3 * * *`; channel rows (telegram/discord/slack) get NULL.
+Cron-driven autosync is on by default (`METATRON_AUTOSYNC_ENABLED=true`). See
+`docs/adr/2026-06-09-autosync-architecture.md` for design rationale.
+
 ## Key Config (env vars, prefix METATRON_)
 - AUTH_ENABLED (false) — JWT gate on /api/v1/*
 - LLM_PROVIDER (ollama) — ollama|deepseek|openrouter|custom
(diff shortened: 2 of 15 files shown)


pahomm:
---
  WARNINGs

  1. status='syncing' can get permanently stuck → connection silently never autosyncs again
  claim_connection_for_autosync requires status != 'syncing'. The claim sets status='syncing'; _run_connection_sync's finally resets it to active/error (connections.py:912) — good on the normal path. But the spawned
  _run_connection_sync tasks live in self._inflight and are not awaited or cancelled on shutdown. On lifespan shutdown only autosync_task (the loop) is cancelled (app.py:236-240); the in-flight child syncs are orphaned and die with
  the loop before their finally runs. Same risk on a hard crash. There is no reaper for stale 'syncing' rows (confirmed: no such query in postgres.py, and freshness's reclaim pass doesn't cover connections).

  Consequence: one ill-timed restart mid-sync leaves a connector pinned at 'syncing' forever — and unlike the manual path (which surfaces a visible 409), autosync just silently skips it indefinitely.

  Suggestions (pick one):
  - On shutdown, also cancel/await scheduler._inflight tasks (the scheduler already tracks them).
  - Add a startup or periodic reaper that resets status='syncing' rows whose sync_log is stale (e.g. running older than N minutes) back to active.

  2. Tautological tests give false coverage for the actual SQL filter logic
  TestListDueAutosyncConnections and TestClaimConnectionForAutosync mock the very method under test, then assert the mock's return value (test_autosync.py:1150-1184, 1080-1098). E.g. test_disabled_connection_excluded does
  store.list_due_autosync_connections = AsyncMock(return_value=[]) then assert due == [] — it verifies the mock, not the WHERE enabled = true AND sync_cron IS NOT NULL AND status != 'syncing' AND (next_run_at IS NULL OR next_run_at
  <= now()) clause, which is where the real logic lives and currently has zero coverage. test_null_coalesces_to_single_run likewise just drains a side_effect.

  The tick() tests (TestAutoSyncSchedulerTick) are the good ones — they exercise real _process_due_row logic (channel-skip, capacity, claim-lost). Recommend converting the SQL-filter tests to integration tests against a real PG (the
  repo has make test-all/integration tier), or at minimum delete the tautological ones so they don't read as coverage they don't provide.

  3. Inconsistent initial-sync behavior: new connections wait, backfilled ones fire immediately
  On create, _validate_and_next_run("0 3 * * *") stamps next_run_at at the next 03:00 (future) — so a freshly created connector does not autosync until 03:00. The migration leaves next_run_at = NULL for backfilled existing rows →
  "due now" → they fire on the first tick after deploy. This asymmetry isn't called out in the ADR. If an initial sync on creation is desired, set next_run_at = NULL (or now()) on create; if not, the divergence is at least worth a
  one-line note. (Manual /sync exists as a workaround, but the inconsistency is surprising.)

  4. Failed/partial sync waits a full cron cycle to retry
  next_run_at is advanced at claim time, before the sync runs (by design per the ADR). A failed sync sets status='error' and next_run_at=tomorrow 03:00, so there's no retry until the next nightly slot. Acceptable for nightly
  cadence, but for short crons (*/15) a transient failure costs a full interval. Worth confirming this is intended; a "retry sooner on failure" path may be a follow-up ticket.

  ---
  SUGGESTIONs

  - Duplicated cron→next-run logic. _validate_and_next_run (connections.py:664) and _process_due_row's next-run computation (autosync.py:378-397) implement the same "croniter in tz → astimezone(UTC)" with the same tz-fallback.
  Extract a shared helper to keep the two in lockstep.
  - Hardcoded "0 3 * * *" in the route (connections.py:569-571) duplicates the migration's server_default. If the default ever changes, the route silently diverges. Pull it into a module constant (e.g. DEFAULT_SYNC_CRON) shared by
  config/route.
  - INFO-level logging on every claim/schedule write. claim_connection_for_autosync (postgres.py:835) and set_connection_schedule log at INFO on every call. With a 60s poll and many due rows at 03:00, that's steady noise — debug is
  more appropriate for the per-row claim.
  - tzdata not declared. zoneinfo.ZoneInfo("Europe/Amsterdam") raises on Windows/minimal containers without the tzdata package; the code silently falls back to UTC, so a configured non-UTC timezone would be ignored with only a
  warning. "UTC" works without it (default is fine). If non-UTC timezones are meant to be supported on all deploy targets, add tzdata to dependencies.
  - Lazy imports inside _validate_and_next_run (from croniter import croniter, import zoneinfo) are inconsistent with autosync.py importing both at module top. Minor; hoist for consistency unless the lazy import is intentional.
  - list_due_autosync_connections is intentionally not workspace-scoped — this is a documented, justified exception to the project's "all queries filter by workspace_id" rule (isolation enforced downstream in _run_connection_sync).
  Good that it's flagged in both the docstring and ADR; no change needed, just noting it for the reviewer record.

  ---

PR #138, opened 2026-06-10
feat(MTRNIX-397): fix RAG core retrieval — query-level scoring + channel triggers

pahomm:
## Summary
Fixes the two core RAG blockers found in trace review (channels dead + degenerate scoring), plus secondary defects. All new behavior is **flag-gated; defaults = prior behavior**.

### Blocker 2 — scoring
- **S1**: query-level signal normalization — `compute_signal_score(active_channels=…)` normalizes only over channels that actually returned results, so the score no longer collapses to `source_balance` noise when metadata/graph are empty. Default-on (inert when all weighted channels return results).
- **S2/S3**: non-zero `dense`/`recency` floors for execution/temporal (interim; per-profile sum kept 0.85). Final weights to be re-derived by grid-search on a corpus-matched eval set (deferred — see below).

### Blocker 1 — recall channels
- **B0**: FAST-LLM slot extraction (dates/people/jira-keys/entities/activity) unioned onto the regex triggers. Hardened: flag + timeout + strict JSON + regex fallback.
- **M-resolver / M7**: resolver emits ISO date **ranges**; ranges carried through recall.
- **M3**: task-tracker activity vocabulary (sprint/task/backlog/…).
- **G1 (bug)**: graph channel seeded from extracted entities instead of the dead `get_graph_entities(query)` exact-text-match path.
- **M5a / M6**: index Jira `duedate` → `due_date` payload; `search_by_date(key=)`; future-date → recent-in-progress fallback.

### Other
- **FAST/HEAVY DeepSeek tier** by `call_site` (`llm/tiers.py`): service calls → FAST, `rag_answer` → HEAVY. Empty `DEEPSEEK_MODEL_FAST` inherits heavy.
- **H1**: non-destructive anti-hallucination guard (flags ungrounded ticket refs) + answer-prompt rule.
- **logging**: force UTF-8 on stdout/stderr (kills Windows cp1251 `UnicodeEncodeError` on Unicode log content).

## Validation
Live on the MTRNIX stand: regression queries ("what are next weeks tasks", "check jira sprint for week 22", "jira+github next week activity") now return **grounded next-week Jira tasks** (with assignees/due dates) instead of digest-dominated / hallucinated answers; metadata channel fires 0→10, all 4 channels fire on multi-entity queries, `ungrounded_tickets: []`. The previously-working "what model do you use" query did not regress. New unit tests for all four mechanisms green.

## Flags (defaults = today)
`METATRON_RETRIEVAL_SCORING_NORMALIZE_ACTIVE_ONLY` (true), `…_SLOT_EXTRACTION_ENABLED` (false), `…_GRAPH_NER_ENABLED` (false), `…_FUTURE_DATE_FALLBACK_ENABLED` (false), `DEEPSEEK_MODEL_FAST` (empty=inherit). Deploy env: mtrnix/infra#4.

## Deferred (separate tickets)
- Phase 6.1 grid-search re-run — needs a corpus-matched, temporal-augmented labeled eval set.
- `workspaces.description` UndefinedColumn (breaks LLM telemetry) — pre-existing, separate.
- non-RAG bypass (2.3, product sign-off), H2 dedup / ingestion double-index, M5b sprint customfield, P1 source=memory digests.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -93,8 +93,13 @@ OLLAMA_NUM_CTX=32768                  # context window for Ollama chat
 OLLAMA_REQUEST_TIMEOUT=               # seconds; empty = library default
 
 # --- DeepSeek ---
+# Active models: deepseek-v4-pro (capable) / deepseek-v4-flash (fast, cheap).
+# deepseek-chat / deepseek-reasoner are DEPRECATED (retire 2026-07-24) — avoid for new setups.
 DEEPSEEK_API_KEY=
-DEEPSEEK_MODEL=deepseek-chat
+DEEPSEEK_MODEL=deepseek-v4-pro         # HEAVY tier (final RAG answer)
+# MTRNIX-397: FAST tier for service LLM calls (resolve/expand/classify/NER/slots).
+# Empty = inherit DEEPSEEK_MODEL at runtime. Only used by the deepseek provider.
+DEEPSEEK_MODEL_FAST=deepseek-v4-flash
 
 # --- OpenRouter ---
 OPENROUTER_API_KEY=
@@ -145,6 +150,19 @@ HYDE_ENABLED=false                    # HyDE for short/vague queries
 ADAPTIVE_RRF_ENABLED=false            # adaptive RRF fusion (regresses metrics, off)
 ENABLE_SEMANTIC_MATCHING=true         # semantic entity matching in resolver
 METATRON_HIERARCHICAL_CHUNKING_ENABLED=true
+
+# --- MTRNIX-397 retrieval core fixes (defaults = today's behaviour) ---
+# S1: normalize the signal score by the channels that actually returned results (query-level).
+# Inert when every weighted channel returns >=1 result; safe on by default.
+METATRON_RETRIEVAL_SCORING_NORMALIZE_ACTIVE_ONLY=true
+# B0: FAST-LLM slot extraction (dates/people/jira keys/entities/activity) feeds channel
+# triggers on top of regex. Off = regex-only (unchanged). Hardened: timeout + JSON + fallback.
+METATRON_RETRIEVAL_SLOT_EXTRACTION_ENABLED=false
+METATRON_RETRIEVAL_SLOT_EXTRACTION_TIMEOUT=6   # slot-extraction LLM timeout (seconds)
+# G1: seed the graph channel from extracted entities instead of the dead exact-text NER path.
+METATRON_RETRIEVAL_GRAPH_NER_ENABLED=false
+# M6: when a resolved date is beyond the corpus, fall back to recent in-progress items.
+METATRON_RETRIEVAL_FUTURE_DATE_FALLBACK_ENABLED=false
 
 # --- HyDE tuning ---
 HYDE_MAX_WORDS=4                      # max query word count to trigger HyDE
--- a/scripts/rag_eval_397.py
+++ b/scripts/rag_eval_397.py
@@ -1,0 +1,162 @@
+#!/usr/bin/env python3
+"""MTRNIX-397 reproducible RAG eval — before/after harness.
+
+Runs a fixed set of queries (three buckets: regression / positive / adversarial) through
+the live REST chat endpoint, pulls each RAG debug trace, and prints a structured per-query
+summary so a reviewer can diff "before" vs "after" without eyeballing raw traces.
+
+This is the M-4 merge gate for the regression bucket. It needs a live stand (Postgres /
+Qdrant / Neo4j) and trace capture enabled (METATRON_RAG_TRACE_ENABLED=true). It is NOT a
+unit test — run it manually:
+
+    python scripts/rag_eval_397.py \
+        --base-url http://localhost:8000 \
+        --email "$METATRON_ADMIN_EMAIL" --password "$METATRON_ADMIN_PASSWORD" \
+        --workspace MTRNIX --out before.json
+
+Re-run after each phase with --out after.json, then diff the two JSON files.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import re
+import sys
+import time
+import urllib.request
+from collections import Counter
+from typing import Any
+
+# Query buckets -------------------------------------------------------------
+
+REGRESSION = [
+    "what are next weeks tasks",
+    "Check jira sprint for week 22",
+    "Use jira mtrnix and GitHub to show next week activity",
+    "what type of model do you use? how many parameters?",  # must NOT regress
+]
+POSITIVE = [
+    "what is MTRNIX-281 about",
+    "summarize the 2026-05-08 daily summary",
+    "who works on the freshness pipeline",
+    "what was done for memory health",
+]
+ADVERSARIAL = [
+    "what are the tasks for the week of 2027-01-04",  # future, no data → must not fabricate
+    "extract the link from https://example.com/auth?code=abc",  # non-RAG
+    "что в спринте на следующей неделе и кто над чем работает",  # cross-language, multi-intent
+    "tell me about the Atlantis migration project",  # likely absent → must say "not found"
+]
+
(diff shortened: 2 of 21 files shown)


pahomm:
Thanks — all six addressed in the amended commit:

- **[WARNING] M6 flag semantics** — added a `max(date_list) > today (UTC)` guard before the fallback in both `recall_metadata`/`recall_metadata_async`, so it only fires for genuinely **future**-dated queries (historical "what happened on 2024-01-01" with an empty week no longer triggers it). Name now matches behavior.
- **[WARNING] find_ungrounded_tickets substring** — switched to exact ticket-key **set membership**: `grounded = {k.upper() for k in _JIRA_KEY_RE.findall(" ".join(parts))}`. Added a regression test (cited `MTRNIX-2` is flagged even when sources contain `MTRNIX-281`). Trace metric no longer lies.
- **[WARNING] M5a needs re-sync** — documented in the infra deploy checklist (mtrnix/infra#4 comment): M5a is inert until the Jira connector is re-synced to populate the `due_date` payload.
- **[SUGGESTION] UTC** — `extract_slots` now uses `datetime.now(UTC)` (aligns the week boundary with `recency_score`).
- **[SUGGESTION] needs_retrieval** — added a comment noting it's reserved for the deferred non-RAG bypass (2.3) and not yet consumed.
- **[SUGGESTION] eval script** — replaced the internal hostname + admin creds with `http://localhost:8000` and `$METATRON_ADMIN_EMAIL`/`$METATRON_ADMIN_PASSWORD` placeholders.

ruff clean, targeted tests green (incl. new substring guard test).

godedok:
## Code Review — MTRNIX-397: RAG core retrieval fixes

Thorough, well-documented PR. Every new behavior is flag-gated with `None`/legacy fallbacks, the math in `compute_signal_score` is correct and well-tested, and the tiering/slot-extraction/ungrounded-ticket helpers all have solid unit coverage. A few things to address before merge.

### 🔴 Blocker

**1. `recall_metadata` sync vs async diverge on the M6 future-date fallback — and the async path is the live one.**

`src/metatron/retrieval/channels.py`

The **sync** version guards the fallback so it only fires for *future* dates:
```python
elif (
    getattr(ctx.settings, "retrieval_future_date_fallback_enabled", False) is True
    and not date_hits
    and date_list
    and max(date_list) > datetime.now(UTC).strftime("%Y-%m-%d")
):
```
The **async** version (`recall_metadata_async`) drops the future-date check entirely:
```python
elif (
    getattr(ctx.settings, "retrieval_future_date_fallback_enabled", False) is True
    and not date_hits
):
```
Since `hybrid_search_and_answer` is async, the async branch is what runs in production. The result: with `METATRON_RETRIEVAL_FUTURE_DATE_FALLBACK_ENABLED=true`, a **historical** dated query that returns no hits (e.g. "what happened on 2024-01-01" with an empty day) will incorrectly fall back to recent in-progress items — exactly the case the sync comment says it must NOT trigger ("historical ... does NOT trigger the fallback"). The async docstring even acknowledges the intent ("a future week with no docs yet") but the code doesn't enforce it. Please mirror the `date_list and max(date_list) > today` guard into the async version. Impact is contained today (flag defaults off), but it's a latent correctness bug on the production path.

### 🟡 Warnings

**2. `execution`/`temporal` profile weight changes (S2/S3) are NOT flag-gated.**

`src/metatron/retrieval/query_classifier.py`

Unlike the rest of the PR, these weight changes ship unconditionally — they alter scoring for every execution/temporal query the moment this merges, regardless of any flag. The "defaults = today's behaviour" framing in the PR description doesn't hold for these two profiles. The values are explicitly interim ("to be replaced by a grid-search re-run") and are validated only by live-stand observation, not by `make grid-search`/eval. That's a real regression risk for query types outside the three regression queries. Recommend (a) calling this out explicitly as an intentional unconditional change, and (b) running `make eval-compare` against the labeled set before merge to confirm no regression on documentation/relationship/user_file profiles.

**3. `due_date` requires a Jira re-sync to take effect.**

`connectors/jira.py` / `jira_processing.py` index `duedate` → `due_date` payload only at ingestion time. Existing Qdrant documents won't have the field until re-synced, so the M5a due-date channel is silently inert on already-indexed corpora. Worth a one-line note in the deploy runbook (re-sync Jira connections after deploy).

**4. No test covers the `recall_metadata` changes (M5a/M6).**

The exact code paths with the sync/async divergence above have zero unit coverage — that's why the bug slipped through. A small test asserting the future-date guard (and that historical empty-date queries do *not* fall back) on the async path would have caught it and would prevent regression.

### 🟢 Suggestions / Notes

- **Pre-existing broken test fixed here:** `test_mixed_matches_current_defaults` on `develop` asserts mixed = `0.35/0.15/0.20/0.10/0.30`, but `develop`'s source already has `0.4/0.0/0.4/0.0/0.6` — i.e. that test is currently failing on `develop`. This PR's rename to `test_mixed_profile_weights` + updated assertions corrects it. Good catch, but worth flagging that it's an unrelated pre-existing breakage being repaired (not introduced by MTRNIX-397).
- **Slot extraction adds a serial LLM hop.** In `hybrid_search_and_answer`, `extract_slots` is awaited before `expand_query`; the two are independent and could run concurrently via `asyncio.gather` to avoid stacking two FAST-LLM round-trips on the critical path when the flag is on. Minor, flag is off by default.
- **`needs_retrieval` is parsed but unused.** Documented as reserved for the deferred non-RAG bypass — fine, just confirm it's intentional dead-until-R1 and not an accidental drop.
- **`logging.py` global `reconfigure`** affects all stdout/stderr process-wide, not just Windows. `errors="replace"` makes it benign, and it's wrapped in `suppress(Exception)`, so acceptable — just noting it's a process-wide side effect of `configure_logging`.
- **Lateral L2↔L2 import:** `search.py` importing `get_dates_in_range` from `ingestion.processors.dates` continues an existing pattern (same layer), so no layer violation — noting for awareness.

### Verdict
One blocker (#1, sync/async fallback divergence on the production path) and one merge-gating recommendation (#2, run `make eval-compare` for the unconditional weight changes). Everything else is solid. Resolve #1 and confirm eval shows no regression, and this is good to go.

pahomm:
Second-round review — addressed:

**1. [BLOCKER] sync/async M6 divergence — you're right, my fault.** The earlier `replace_all` only matched the sync branch (the two had different comments), so `recall_metadata_async` (the production path) shipped without the future-date guard. Fixed: the async branch now has the identical `date_list and max(date_list) > today(UTC)` guard. Historical empty-date queries no longer fall back.

**4. [WARNING] no test on recall_metadata M5a/M6** — added `tests/unit/retrieval/test_recall_metadata_future_fallback.py` on the **async** path: future-date+no-hits → fallback fires; historical-date+no-hits → does NOT fall back; flag-off → never. This reproduces/locks the bug from #1.

**2. [WARNING] S2/S3 weights unconditional** — acknowledged, and you're right the 'defaults = today' framing doesn't cover these. Scoping note: only **execution** and **temporal** presets changed; `documentation`/`user_file`/`relationship`/`mixed` are byte-identical, so the regression surface is exactly the two profiles we targeted and validated live. They're interim (grid-search re-run deferred). Re `make eval-compare`: can't run meaningfully right now — the labeled testset's `expected_doc_labels` are tied to the full corpus (Confluence page IDs + uploads), but this stand is freshly re-synced **Jira-only**, so eval would score documentation/file queries as total misses. That corpus-parity gap is exactly why grid-search/eval is deferred (noted in PR + ticket).

**3. due_date re-sync** — documented in the infra deploy checklist (mtrnix/infra#4 comment): trigger a Jira re-sync after deploy; M5a is inert until `due_date` payload exists.

**Green notes:** pre-existing `test_mixed` breakage — yes, repaired here, not introduced (flagged in the test comment). `needs_retrieval` — intentional, reserved for deferred 2.3 (now commented). `extract_slots`/`expand_query` concurrency — fair, noted as an optional follow-up (flag off by default, keeping this PR focused on correctness). logging process-wide reconfigure + lateral L2 dates import — acknowledged, both intentional/benign.

ruff clean; new async M6 test + recall_channels green.

PR #211, opened 2026-06-25
Improve installer Neo4j startup guidance

toomij99:
## Summary
- print actionable guidance when Docker Compose fails during installer launch, especially for Neo4j first-start password mismatches
- cover non-interactive resume behavior, secret generation failure, and Compose launch failure messaging in installer tests

## Tests
- bash -n install.sh
- bash tests/installer/test_resume.sh


toomij99:
Closing this duplicate because this branch should target develop. Active PR: #210.

PR #269, opened 2026-07-01
chore: retire the develop branch — point everything at main

godedok:
Prep for consolidating onto a single long-lived branch (`main`). Updates every reference to `develop` so nothing breaks when the branch is deleted.

## Changes

- **README.md / install.md** — `git clone -b develop …` → plain `git clone …` (clones the default branch, now `main`). This is the one that would otherwise **break onboarding**: `-b develop` fails once the branch is gone.
- **.github/dependabot.yml** — `target-branch: "develop"` → `"main"` for both the `uv` and `github-actions` ecosystems (otherwise Dependabot targets a non-existent branch).
- **.github/workflows/lint.yml** — push trigger `[ "develop", "main" ]` → `[ "main" ]`.
- **CONTRIBUTING.md** — "Branch from `develop`" / "open PRs against `develop`" → `main`.
- **SECURITY.md** — supported-versions table row `develop (latest)` → `main (latest)`.

## Out of scope (intentionally left)

- The placeholder secret string `develop-secret-key-change-in-prod` (not a branch).
- A docker image tag `splade-service-develop` and a `seed/` URL pointing at the separate `metronixcore` repo — unrelated to this repo's branch.

## After this merges

Delete the `develop` branch. It's already fully merged into `main`, so no history is lost.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,7 +3,7 @@ updates:
   # Python dependencies (uv lockfile + pyproject).
   - package-ecosystem: "uv"
     directory: "/"
-    target-branch: "develop"
+    target-branch: "main"
     schedule:
       interval: "weekly"
     # Paused during OSS launch: 0 stops new version-update PRs while still
@@ -23,7 +23,7 @@ updates:
   # GitHub Actions versions used in workflows.
   - package-ecosystem: "github-actions"
     directory: "/"
-    target-branch: "develop"
+    target-branch: "main"
     schedule:
       interval: "weekly"
     # Paused during OSS launch (see note above). Restore to 5 to resume.
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -3,7 +3,7 @@ name: Lint
 on:
   pull_request:
   push:
-    branches: [ "develop", "main" ]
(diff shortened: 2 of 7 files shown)


pahomm:
The docstring contains "...these are the grid-searched values committed on develop..." — this is the only remaining reference to the develop branch inside the repository itself that this PR does not update. The PR's stated goal is to "update every reference to develop so nothing breaks." Once the branch is deleted, this comment becomes a dangling reference to a branch that no longer exists.

Failure scenario: A future reader sees "committed on develop," tries to inspect that branch for context, and finds it gone — mild confusion, no functional breakage (the weight values themselves remain valid). Recommend rewording for consistency with the PR's stated goal, e.g. "committed historically" or pointing to a specific commit/tag.

Sergey-Vasnev (dismissed):

godedok:
Good catch — you're right, that one slipped through the doc/config sweep. Reworded the docstring in `tests/unit/test_query_classifier.py` from "grid-searched values committed on develop" to "committed grid-searched values", so no reference to the branch remains anywhere in the repo. Comment-only — the weight values and assertions are unchanged, and the test still passes. (`fae709d`)

pahomm (approved):

PR #270, opened 2026-07-01
feat: native OpenClaw integration (mirrors Hermes wiring)

### LINKED ISSUE #266: feat/ native integration with openclaw
### Problem / motivation

What:

We need a native integration path for openclaw: it presents in installation path, installation script

Why:

This native integration will help user more convenient integrate this tool with his openclaw

### Proposed solution

- install.sh has auto research and finding of the openclaw files;
- installation guide and path describes the manual integration with openclaw

### Alternatives considered

_No response_

godedok:
## Summary

Closes #266. Gives `install.sh` a deterministic, no-LLM wiring path for OpenClaw (`wire_openclaw`), matching the existing Hermes path (`wire_hermes`), plus two related fixes found along the way.

- **OpenClaw wiring**: detects OpenClaw (CLI binary and/or `~/.openclaw`), registers Metronix as an MCP server via OpenClaw's own `openclaw mcp show`/`openclaw mcp set` CLI (never hand-parses the JSON5 config), appends an availability note to `SOUL.md`, and falls back to the existing paste-ready prompt guide on any detection gap or CLI failure.
- **New `--wire-openclaw` flag**, wired into the interactive `connect_agent()` menu (now Hermes / OpenClaw / other) and a standalone `main()` shortcut mirroring the existing Hermes one.
- **Bug fix**: `_backup_file()` returns nonzero when its target doesn't exist; calling it bare (no `|| true`) as the last statement before a `set -e` script's next command aborts the whole installer. This could hit a real Hermes install where `~/.hermes/` exists but `config.yaml`/`SOUL.md` don't yet. Fixed in `wire_hermes` and applied proactively to the new OpenClaw path, with a regression test.
- **Unified** the prompt-output directory: Hermes and OpenClaw both now write to `metronix-agent-setup/` (previously Hermes used its own `metronix-hermes-setup/`).
- **Docs**: rewrote `docs/integrations/openclaw.md` with concrete paths/commands and a confidence caveat (the CLI schema was sourced from `docs.openclaw.ai`, not verified against a live install — though cross-checked against a real, older local install during development); filled in `connecting_to_agent.md`'s file-location table.

Two ideas were investigated and explicitly rejected (not implemented): `.env`/`${VAR}`-style credential indirection for either runtime (both have relevant open upstream bugs — OpenClaw's HTTP MCP headers don't expand `${VAR}`, Hermes config migration can bake a placeholder back to a literal), and switching Hermes's existing config edit to `hermes mcp add` (undocumented flag surface for the HTTP+headers case needed here).

## Test plan

- [x] `shellcheck install.sh` — clean
- [x] `bash -n install.sh` — clean
- [x] `tests/installer/test_wire_openclaw.sh` — 35/35 passing (new test file, fake `openclaw` CLI stub, covers detection, idempotent registration, CLI-failure fallback, menu/standalone wiring)
- [x] `tests/installer/test_wire_hermes.sh` — 57/57 passing (54 pre-existing + 3 new for the `_backup_file` regression)
- [x] `tests/installer/test_resume.sh`, `test_install.sh`, `test_failgen.sh` — unaffected, all passing
- [x] Manual smoke test with a fake `openclaw` on `PATH`


godedok:
Marking this as **draft** — the automated test suite passes (see test plan above), but this needs thorough manual local verification before it's ready for review:

- Fresh local install with `./install.sh --wire-openclaw -y` against a real OpenClaw installation
- Interactive `connect_agent()` menu path (option 2)
- Manual sanity check of the resulting `~/.openclaw/openclaw.json` and `~/.openclaw/workspace/SOUL.md` content

Will move out of draft once that's done.

toomij99 (changes requested):
Please, think about naming of integration or connect flag/functions. wire -> setup/install/register/configure/enable or last option "connect"

toomij99 (dismissed):
Thank you

godedok:
Manual local verification completed — moving out of draft:

- Wired a real OpenClaw 2026.6.11 install via `./install.sh --connect-openclaw -y`: MCP entry written through `openclaw mcp set`, config validates, `openclaw mcp probe metronix` reports 22 tools.
- End-to-end confirmed from a live agent session: `metronix_status(workspace_id="MTRNIX")` answered over MCP.
- Old-CLI fallback confirmed on 2026.3.13 (no `mcp` subcommand): installer degrades to the paste-ready guide without touching any files.
- Testing surfaced one real defect, fixed in ba00689: re-running the connect step after a stack reinstall (rotated `METRONIX_MCP_API_KEY`) used to keep the stale key while reporting "already present"; the state check now compares URL + key + agent id and re-registers on mismatch.

Known follow-up (separate issue): the same stale-key idempotency gap exists in the Hermes/Claude Code/Codex connect flows inherited from main.

pahomm:
Findings

1. install.sh:776-778 & install.sh:1142-1146 — the _backup_file guard rationale contradicts the actual (and newly-tested) behavior.
Both new comment blocks state _backup_file "returns 1 [when the target] doesn't exist... which would otherwise trip set -e", justifying the added _backup_file … || true guards. But _backup_file uses the if [[ -f "$1" ]]; then cp …; fi form on both main and this branch, which returns 0 on a missing file (verified: a bare call under set -euo pipefail on a nonexistent path exits 0 and continues). This PR's own new tests/installer/test_backup_file.sh asserts exactly that ("must never return nonzero on a missing target"). So within one PR the test says "returns 0" while the guard comments say "returns 1." Net effect: the || true additions are harmless but dead/redundant, and the inline comments actively misdescribe the code. A maintainer should either drop the guards+comments or, if they want defense-in-depth, reword the comments to "defensive — _backup_file is contractually 0 (see test_backup_file.sh)". (Not a runtime bug — correctness-of-rationale / maintainability.)

2. install.sh:1062-1072 (openclaw_mcp_state) — idempotency/stale-key detection assumes openclaw mcp show prints the API key in cleartext. has_current requires grep -qF "$H_KEY" to match the mcp show output. Credential-bearing CLIs commonly redact secrets in show/get output OpenClaw does, the key never matches, has_different is alwa set is re-invoked on every run — the exact idempotency the
PR #270 adds a deterministic, no-LLM OpenClaw wiring path to install.sh (connect_openclaw), mirroring the existing Hermes flow: it detects OpenClaw, registers Metronix as an MCP server via OpenClaw's own openclaw mcp set CLI (never hand-editing the JSON5 config), merges an availability block int~/.openclaw/workspace/SOUL.md, and falls back to the paste-. It also refactors the three duplicate write_*_prompt_dirfunctions into one write_runtime_prompt_dir, unifies the setup-prompt directory naming, and updates the install/uninstall/integration docs. The refactor is clean and the new code is well-tested (a fake openclaw CLI stub exercises detection, idempotency, key rotation, and CLI-failure fallback).
                                                                                                                                                   Findings
                                                                                                                                                   1. install.sh:776-778 & install.sh:1142-1146 — the _backup_cts the actual (and newly-tested) behavior.
Both new comment blocks state _backup_file "returns 1 [when the target] doesn't exist... which would otherwise trip set -e", justifying the added  _backup_file … || true guards. But _backup_file uses the if form on both main and this branch, which returns 0 on amissing file (verified: a bare call under set -euo pipefail on a nonexistent path exits 0 and continues). This PR's own new                        tests/installer/test_backup_file.sh asserts exactly that ("a missing target"). So within one PR the test says "returns0" while the guard comments say "returns 1." Net effect: the || true additions are harmless but dead/redundant, and the inline comments actively misdescribe the code. A maintainer should either drop the guards+comments or, if they want defense-in-depth, reword the comments to "defensive —   _backup_file is contractually 0 (see test_backup_file.sh)".ness-of-rationale / maintainability.)

2. install.sh:1062-1072 (openclaw_mcp_state) — idempotency/stale-key detection assumes openclaw mcp show prints the API key in cleartext. has_current requires grep -qF "$H_KEY" to match the mcp show output. Credential-bearing CLIs commonly redact secrets in show/get output (e.g. Bearer ****). If OpenClaw does, the key never matches, has_different is alwa set is re-invoked on every run — the exact idempotency thestub test "same key → set not re-invoked" locks in would fail against a real CLI. Functionally harmless (it re-registers identical values), but it defeats the claimed idempotency and the whole point of the he PR itself flags the CLI schema as sourced from docs, notverified live — worth a note that this branch is untested against real openclaw output.                                                            
3. uninstall.md:822 — openclaw mcp unset metronix is an unverified inverse of the install's openclaw mcp set. The install path only ever uses mcp show/mcp set (deliberately conservative about the CLI surface), but the uninstall doc introduces mcp unset, which isn't exercised anywhere. If the real removal subcommand is mcp remove/mcp delete, the documo-ops. Low severity (doc only), but easy to align with aconfidence caveat like the one already in docs/integrations/openclaw.md.

Nothing else survived verification — the prompt-dir refactohortcut, resolve_agent_id optional-arg change(${2:-}/${3:-}), and json_escape are all correct and consistent with the existing Hermes/Codex patterns.

godedok:
Thanks for the review — all three findings addressed in 50882f2 (plus 0527fc9 for the earlier naming request):

**1. `_backup_file` guards / "returns 1" comments — confirmed, fixed.** The guards and comments were written against the pre-merge implementation (`[[ -f ]] && cp`, which genuinely returned 1 on a missing target). The merge from main brought in the `if/fi` form that returns 0, which made the guards dead code and the comments wrong — exactly as you said. Dropped both `|| true` guards and their comments; bare call sites now match the Claude Code/Codex style, and the missing-file-returns-0 contract is documented at the function definition and locked in by `tests/installer/test_backup_file.sh`.

**2. `mcp show` redaction assumption — verified live, now documented in code.** `openclaw mcp show metronix` on a real OpenClaw 2026.6.11 install prints the full entry including the unredacted `Authorization` header (this was checked before implementing the key comparison, but the evidence only lived in a commit message). `openclaw_mcp_state` now carries a comment stating the live verification and the failure mode if a future version redacts: the key grep stops matching and every run re-invokes `mcp set` with identical values — idempotency degrades, but it fails toward harmless re-registration, never toward keeping a stale key.

**3. `openclaw mcp unset` — verified live.** `openclaw mcp --help` on 2026.6.11 lists `unset — Remove one OpenClaw-managed MCP server`, so the uninstall doc is correct. The confidence note in `docs/integrations/openclaw.md` is updated accordingly: the config path, `set`/`show`/`unset`, the JSON payload schema, and the `SOUL.md` workspace path are now all marked as verified against a live 2026.6.11 install rather than docs-sourced only.

**Naming (earlier review):** functions and flags were already `connect_*` after the main merge; 0527fc9 renames the last leftovers — `test_wire_hermes.sh`/`test_wire_openclaw.sh` → `test_connect_hermes.sh`/`test_connect_openclaw.sh`.

All 6 installer test suites pass (194 assertions), shellcheck clean.

pahomm (approved):

PR #273, opened 2026-07-02
feat(connectors): GitHub connector (issues, PRs, docs, releases)

pahomm:
## Summary

Implements the previously scaffolded **GitHub connector** (it was a stub raising `NotImplementedError`), following the existing Jira/Confluence/Notion pattern. Metronix can now ingest GitHub repository content into the knowledge base.

## What it indexes

- **Issues** (open + closed) — body + all comments
- **Pull requests** (open + closed) — body + issue comments + review comments
- **README** + repository **Markdown docs** (`*.md`, ≤1 MB)
- **Releases** (changelog bodies)

All emitted as `source_type=github` with stable `source_id` schemes (`gh-issue-*`, `gh-pr-*`, `gh-readme-*`, `gh-doc-*`, `gh-release-*`).

## Configuration

| Key | Required | Purpose |
| --- | --- | --- |
| `token` | yes | Personal Access Token |
| `org` | no | Organization / owner |
| `repos` | no | `repo1,repo2`, or empty for all accessible repos |
| `base_url` | no | GitHub Enterprise Server API base |

## Design notes

- **Incremental sync**: issues use the server-side `since` filter; PRs iterate newest-first and stop at the cursor; README/docs/releases are re-fetched and de-duplicated by content hash in the sync layer.
- **Off-loop I/O**: PyGithub is synchronous with lazy `PaginatedList`s — each repo's whole unit of work runs inside a single `asyncio.to_thread`.
- **Rate limits**: rely on PyGithub's built-in `retry`.
- `source_role = "knowledge_base"`.
- Metadata is kept `dict[str, str]` (lists → CSV, bools → `"true"/"false"`).

## Review fixes folded in

- Resilient repo resolution — a single bad/typo'd repo no longer aborts the whole sync (per-repo try + skip invalid bare names).
- README de-dup — the rendered README (any path: root, `docs/`, `.github/`) is not re-indexed from the tree.
- Warn on truncated git tree instead of silently dropping files.
- `health_check` uses `get_rate_limit()` (valid for repo-scoped tokens) instead of `get_user()`.

## Testing

- New unit suites: `tests/unit/test_github_processing.py`, `tests/unit/test_github_connector.py` (pure processing, config/health, repo resolution, incremental boundary, review-fix regressions).
- 22 tests pass; `ruff` clean.
- Manually verified end-to-end against a live stack: issues/PRs/docs are fetched and persisted (graph-extraction failures during that run were an unrelated local `.env` LLM-key misconfiguration, not the connector).

## Out of scope

Wiki (no REST API — separate `.wiki.git`), GitHub App auth, non-Markdown source files, Discussions, webhook-driven sync.

## Also included

A small fix to the `install.sh` ASCII banner alignment.


toomij99 (commented):
thank you

godedok:
## Review

Nice implementation overall — the connector follows the established Jira/Confluence/Notion pattern well, and the resilience fixes (per-repo resolution, README de-dup, truncated-tree warning) are good calls. Found a few issues worth addressing before merge, ranked by severity. All verified against the actual PyGithub 2.9.1 source and the surrounding codebase.

1. **The connector is unreachable through its own primary interface.** `src/metronix/mcp/tools/source_sync.py:17` still lists `"github"` in `_SCAFFOLD_CONNECTORS`, and `metronix_source_sync` raises `ValueError("Connector 'github' is not implemented yet...")` before `run_connection_sync`/`GitHubConnector.fetch()` is ever called (lines 57-61). An agent can create a GitHub connection but can never sync it via MCP. This PR needs to remove `"github"` from that frozenset.

2. **`source_role: "task_tracker" → "knowledge_base"` silently demotes GitHub evidence for existing and future installs.** `retrieval/search.py`'s `PROFILE_PRIMARY_ROLE` marks `task_tracker` PRIMARY for `execution`/`temporal` queries (e.g. "what's the status of the login bug"). After this change, GitHub issues/PRs lose PRIMARY status for exactly those query types on their next re-sync — no migration, no test covering the interaction. Worth confirming this role change is intentional and documenting the retrieval-side impact.

3. **`metadata["type"]` values (`"github_issue"`/`"github_pr"`/`"github_release"`/`"github_doc"`) break the type==connector_type convention** every other connector follows (jira/confluence/notion all set `metadata["type"]` to their own connector name). `ingestion/pipeline.py` spreads `doc.metadata` after setting `"type"` from the connector type, so GitHub's value wins. `retrieval/search.py`'s `_result_type()` and `retrieval/trace.py`'s source lookup both read this field expecting the connector name, so GitHub content gets the wrong label and feeds `type_counts`/`source_balance` scoring incorrectly.

4. **An expired/revoked token gets reported as a successful, empty sync.** In the explicit `repos=` config path, `_resolve_repos()`'s per-name `try/except` (github.py ~182-186) swallows auth failures from `get_repo()` into warning logs, so `fetch()` returns `[]` without raising. `connection_sync.py` treats that as `status="success"` and advances `last_synced_at`. `jira.py`'s fetch loop re-raises non-429 errors instead, which is the safer pattern here too.

5. **`_fetch_files`/`_fetch_releases` ignore `since` entirely**, so every sync re-walks the whole git tree, re-downloads every markdown file, and re-lists every release regardless of what changed. The PR description says this is covered by content-hash de-dup "in the sync layer," but that dedup path (`check_and_version_document`) isn't actually wired into production ingestion — it's only exercised in tests. Might be fine for small repos, but worth being explicit that there's currently no dedup safety net for this cost.

6. **`configure()` passes `"retry": 3` (a bare int) to `Github()`**, which resolves through `urllib3.Retry.from_int(3)` with an empty `status_forcelist` — meaning **zero** HTTP-status-based retries (no 403/429/5xx backoff at all). PyGithub's own default (`GithubRetry()`) handles exactly this with rate-limit-aware backoff. This is weaker than doing nothing and contradicts the PR description's claim to rely on PyGithub's built-in retry. Suggest just not passing `retry` at all (or passing a real `GithubRetry(...)` instance) to keep the library default.

7. **`getattr(issue, "pull_request", None)` and `getattr(pr, "merged", False)` each trigger an extra full-object GET per item.** List-endpoint responses omit these keys for plain issues/PRs, and PyGithub objects built from a list page are never marked "completed," so accessing an absent attribute triggers a lazy-completion fetch — effectively doubling API calls for issue/PR-heavy repos. Might be worth checking `issue.raw_data.get("pull_request")` / `pr.raw_data.get("merged_at")` directly instead to avoid the round-trip.

8. **Stale docs**: `src/metronix/mcp/tools/source_schemas.py:17-20` and `source_create.py:16-17` still say "Working connectors: confluence, jira, notion... github ... NOT implemented — sync will fail." Combined with #1, this actively steers agents away from a connector this PR ships. Worth updating alongside the `_SCAFFOLD_CONNECTORS` fix.

Not flagged as blocking: the naive/aware datetime mixing in `Document.created_at` (couldn't find a reachable trigger — `since` is always timezone-aware in practice), and the wiki-pages gap (explicitly called out as out of scope in the PR description).

godedok:
## Follow-up review

Re-checked the updated diff. 6 of the 8 previous findings are fixed:

1. ✅ `_SCAFFOLD_CONNECTORS` — `"github"` removed from the frozenset in `source_sync.py`; `source_create.py`/`source_schemas.py` descriptions updated to match; `test_mcp_source_tools.py` updated to test `gdrive` instead. The connector is reachable via MCP now.
2. ✅ `metadata["type"]` — now consistently `"github"` (matches the connector-name convention), with the specific kind moved to a new `"github_type"` field (`issue`/`pull_request`/`release`/`doc`). Clean fix.
3. ✅ Auth-failure-masked-as-success — `_resolve_repos()` now only catches `UnknownObjectException` (404s); everything else propagates. `test_resolve_repos_reraises_non_404` covers this explicitly.
4. ✅ `retry: 3` — removed entirely; falls back to PyGithub's own `GithubRetry()` default, with a comment explaining why the bare int was worse.
5. ✅ Lazy-completion N+1 on `issue.pull_request`/`pr.merged` — now reads `issue._rawData.get("pull_request")` and `pr.merged_at` instead, avoiding the extra per-item GET. Verified against the installed PyGithub source — `_rawData` does hold the original list-page JSON, so this is correct (a nit: the public `.raw_data` property returns the same value and would avoid touching a "private" attribute, but functionally equivalent).
6. ✅ Stale MCP tool docs — fixed alongside #1.

Still open (not addressed, non-blocking):

- `source_role: "task_tracker" → "knowledge_base"` — the retrieval-priority impact (PROFILE_PRIMARY_ROLE demoting GitHub issues/PRs for execution/temporal queries) still isn't documented or tested. May be intentional, but worth an explicit confirmation.
- `_fetch_files`/`_fetch_releases` still ignore `since` — README/docs/releases are still re-fetched in full on every sync.

One new minor nuance from the current fix (checked, not blocking): narrowing `_resolve_repos()`'s except clause to only `UnknownObjectException` means a non-auth transient error on one repo in a multi-repo `repos=` list (e.g. a `GithubException` after retries are exhausted) now aborts resolution of the rest of the list too, which is a step back from the PR's stated "one bad repo shouldn't abort the whole sync" goal — though broadening the catch risks re-introducing the original auth-masking bug, and per-item resilience for the actual data fetch already lives in `_fetch_repo`'s per-repo try/except. Net: a reasonable tradeoff, not something I'd hold up the merge for.

No new bugs found in the delta. LGTM modulo the two open items above, which I'd leave to your judgment on whether to address now or as a fast-follow.

godedok (approved):

Issue #275, opened 2026-07-02
Rename all connect actions during installation from "wire" to "connect"


Sergey-Vasnev:
We have "wire-hermes" but need to rename it to "connect-hermes" as the claude implemented

cycterna2222288888-ai:
Checked against `main` (`1f21741`). The rename looks complete.

Everything uses `connect-` now:
- `install.sh` — `--connect-hermes` flag (`:106`, `:158`), `CONNECT_HERMES` var, `connect_hermes()` (`:700`), `connect_openclaw()` (`:1084`)
- `install.md:51`, `install.md:376`, `README.md:188`
- tests: `tests/installer/test_connect_hermes.sh`, `tests/installer/test_connect_openclaw.sh`

No `wire-hermes` / `wire_hermes` / `wire` action remains anywhere (searched `*.sh`, `*.py`, `*.md`, `Makefile`). The only residual "wire" is prose in comments ("MCP wiring", "un-wireable").

Asking to confirm — can this be closed?

toomij99:
Closing — verified done on `main` @ `1f21741`.

Installer/docs/tests use "connect" rather than "wire" (`install.sh`, `install.md`, `README.md`, `tests/installer/`). Thanks @cycterna for the code check against this issue.

Issue #186, opened 2026-06-23
Develop the Metronix SDK for Claude Code / Codex integration


Sergey-Vasnev:
Implement the SDK that enables Metronix integration with Claude Code / Codex.

toomij99:
Closing as **not planned**.

Claude Code / Codex integration is covered by MCP connect flows (`install.sh --connect-claude/--connect-codex` + integration docs). Separate SDK package not needed. Verified against `main` @ `1f21741` (cycterna audit).

PR #281, opened 2026-07-02
Added codex integration support to installation steps and script. 

Sergey-Vasnev:
## Summary

Adds first-class **Codex** support to the Metronix installer and integration docs, following the same pattern as Hermes and Claude Code.

**Installer (`install.sh`)**
- New `--connect-codex` flag (and interactive menu option 3) to register Metronix by editing `~/.codex/config.toml` with a minimal, validated text insert (backed up first; never a full rewrite)
- Falls back to ready-to-paste prompts in `metronix-codex-setup/` when the config cannot be edited safely
- Extends agent-id resolution to reuse `X-Agent-Id` from existing Codex config (shared `METRONIX_AGENT_ID` anchor in `.env`)
- Non-interactive `-y`: `--connect-codex` selects Codex; otherwise Hermes remains the default

**Documentation**
- Expands `docs/integrations/codex.md` with automatic/manual setup, scope notes, and troubleshooting (including why `codex mcp add` is not the primary path — no `X-Agent-Id` header support)
- Adds four Codex-specific prompt templates under `docs/integrations/codex/` (install, memory policy, migrate, rollback)
- Updates `connecting_to_agent.md`, `install.md`, `prompts.md`, and related guides
- Gitignores `metronix-codex-setup/` (contains MCP API key)

**Naming cleanup**
- Renames `--wire-hermes` → `--connect-hermes` (flag, `CONNECT_HERMES` variable, `connect_hermes()` function) for consistency with `--connect-claude` / `--connect-codex`
- Updates installer tests and docs accordingly

No runtime/API behavior changes — installer, docs, and shell tests only.

## Related issue

https://github.com/mtrnix/metronix-memory/issues/186\
https://github.com/mtrnix/metronix-memory/issues/275

## Type of change

- [ ] Bug fix
- [x] New feature
- [x] Refactor / internal change
- [x] Documentation
- [x] CI / tooling

## Checklist

- [x] `ruff format` and `ruff check` pass locally *(no Python files changed)*
- [x] Tests added/updated and `pytest` passes *(installer shell tests updated for `--connect-hermes`; no Python test impact)*
- [x] Docs updated if behavior or config changed
- [x] Changes are workspace-scoped where applicable (no cross-workspace data access)


godedok:
## Code review

### Overview
This PR adds first-class **Codex** support to the installer (`install.sh`) and docs, mirroring the existing Hermes / Claude Code pattern: a new `--connect-codex` flag (and interactive menu option 3), a minimal validated text edit of `~/.codex/config.toml` (backed up first, never a full rewrite), a fallback to ready-to-paste prompts in `metronix-codex-setup/`, and agent-id resolution extended to read `X-Agent-Id` from an existing Codex config. Alongside it, a mechanical rename `--wire-hermes` → `--connect-hermes` (flag, `CONNECT_HERMES`, `connect_hermes()`) plus a large docs block and 4 prompt templates.

### Result
I traced the diff line by line and checked the `resolve_agent_id`/`resolve_agent_connection` call sites, the `merge_codex_config`/`codex_toml_state` logic, `set -euo pipefail` interactions, rename completeness, and the TOML schema. **No blocking defects found** — the PR is clean, well-commented, and tests/docs are updated in lockstep.

I also verified the Codex `config.toml` keys used (`url`, `http_headers`, `startup_timeout_sec`, `tool_timeout_sec`) against the [official Codex config reference](https://developers.openai.com/codex/config-reference) — they are correct, which cleared the only potentially critical risk.

Things I checked that turned out **not** to be bugs:
- **`_backup_file` under `set -e`** — rewriting `[[ -f ]] && cp` into an explicit `if/fi` (returns 0 when the file doesn't exist yet) is a correct fix for the new case where `connect_codex` calls it on a not-yet-existing `config.toml`, not a regression.
- **Unusual TOML layout** (a pre-existing inline `mcp_servers = { ... }`) would produce a duplicate-key after the append, but the post-edit validation (`toml_read ...metronix.url != H_URL`) catches it and falls back to writing the guide. Handled, with a comment.
- **Rename completeness** — no leftover `wire_hermes` / `WIRE_HERMES` / `--wire-hermes` references remain in `install.sh`; README, `install.md`, `uninstall.md`, `hermes.md`, and the installer tests are all updated.
- **`resolve_agent_id`** — the sole caller is updated to 3 args, and an empty `codex_config` is safely handled by the guard.

### Minor, non-blocking notes
- `install.sh` interactive **project scope** writes `./.codex/config.toml` relative to `cwd`, and `main()` does `cd "$REPO_ROOT"`, so the file lands in the Metronix checkout rather than the user's actual project. This is exactly the same pattern the Claude Code path already uses (`./.mcp.json`), so it's consistent behavior rather than a new defect — worth a doc note at most.
- The `codex mcp list` verification for project scope can emit a false "metronix did not show up" warning, since `codex mcp list` run from `REPO_ROOT` won't read another (untrusted) directory's project config. Cosmetic — it doesn't affect the outcome.

LGTM. 👍

godedok (dismissed):

toomij99 (commented):
[mcp_servers.metronix] table is treated as “already configured” and skipped. That’s risky for the exact migration path this PR is introducing: if someone already added metronix via codex mcp add or an older/manual config, they can easily have a metronix entry that is missing http_headers.X-Agent-Id (or even the auth header). In that case the installer will say “already present,” codex mcp list will still pass because it only checks for presence by name, but the integration is still broken.

Sergey-Vasnev:
> [mcp_servers.metronix] table is treated as “already configured” and skipped. That’s risky for the exact migration path this PR is introducing: if someone already added metronix via codex mcp add or an older/manual config, they can easily have a metronix entry that is missing http_headers.X-Agent-Id (or even the auth header). In that case the installer will say “already present,” codex mcp list will still pass because it only checks for presence by name, but the integration is still broken.

Issue addressed and fixed

godedok (approved):

PR #284, opened 2026-07-03
fix: harden Neo4j/Postgres password lifecycle on reinstall & reset

### LINKED ISSUE #283: Reinstall breaks Neo4j auth: new NEO4J_PASSWORD mismatches existing data volume
## Summary

Reinstalling Metronix intermittently breaks Neo4j authentication: the stack comes up with a Neo4j container that is `unhealthy` (or the API cannot reach the graph), because the newly generated `NEO4J_PASSWORD` no longer matches the password baked into the existing `full_neo4j_data` volume.

## Root cause

Neo4j (like Postgres) only honours `NEO4J_AUTH` / `NEO4J_PASSWORD` on the **first** startup of a data volume. On every later start it keeps the password the volume was initialised with.

On reinstall:

- `install.sh` generates a fresh `NEO4J_PASSWORD` via `gen_secret` when it cannot recover the previous value (e.g. `.env` was removed/regenerated) — see `install.sh:1390` (`resolve_secret NEO4J_PASSWORD "$prev_neo" "$(gen_secret)"`).
- The Docker volume `full_neo4j_data` (`docker-compose.yml:67`, `:244`) persists across the reinstall and still carries the **original** password.
- `docker-compose.yml:65` renders `NEO4J_AUTH: "${NEO4J_AUTH:-neo4j/${NEO4J_PASSWORD:-metronix_dev}}"` from the new `.env`, so Neo4j starts with credentials that no longer match its stored password → `Neo.ClientError.Security.Unauthorized`, container healthcheck (`cypher-shell ... RETURN 1`, `docker-compose.yml:71`) fails, container is `unhealthy`.

The installer already **detects and warns** about this after the fact (`install.sh:1670-1716`, `wait_health` hints at `install.sh:1756-1765`), but it does not **prevent** the mismatch on a normal reinstall — the user still ends up with a broken stack and has to reset the volume manually.

## Steps to reproduce

1. Fresh install: `./install.sh -y` — Neo4j volume `full_neo4j_data` is created with password P1.
2. Simulate a reinstall where the secret is not recoverable: remove or blank `NEO4J_PASSWORD` in `.env` (or delete `.env`) while **keeping** the Docker volume.
3. Run `./install.sh -y` again → new password P2 generated and written to `.env`.
4. Stack starts; `metronix-full-neo4j` becomes `unhealthy`; Neo4j log shows an authentication/`Unauthorized` error. `docker inspect --format='{{.State.Health.Status}}' metronix-full-neo4j` → `unhealthy`.

## Expected

Reinstall should not silently break Neo4j auth. The installer should either:

- detect that a `full_neo4j_data` volume already exists **before** generating a new password and reuse the volume's password (or refuse to overwrite it), or
- offer an explicit, guided path (keep data + restore password, vs. reset volume) instead of only warning after the failed `up`.

## Actual

The stack comes up unhealthy; the only remediation the installer prints is the destructive `docker compose ... down -v` (destroys graph data) or manually restoring the old password (`install.sh:1707-1712`, `1760-1764`).

## Notes / related

- Same class of bug affects Postgres (`warn_if_db_auth_failed`, `install.sh:1725`), which can look healthy while rejecting every query — worth handling consistently.
- Related infra epic: #159.
- Existing installer regression tests live in `tests/installer/` (`test_resume.sh` covers empty `NEO4J_AUTH=` handling) — a regression test for the password/volume mismatch on reinstall should be added with the fix.

## Environment

- Component: `install.sh`, `docker-compose.yml`
- Storage: Neo4j (`full_neo4j_data` volume)

godedok:
Fixes #283

Two related fixes for the DB-password-vs-data-volume lifecycle that caused auth failures on reinstall.

## 1. Refuse a reinstall that would mismatch an existing volume

Neo4j and Postgres fix their password on the **first** start of a data volume and never accept a new one afterwards; the baked-in password cannot be read back. On reinstall, `install.sh` generated a fresh random `NEO4J_PASSWORD` / `POSTGRES_PASSWORD` whenever it could not recover the previous value (e.g. `.env` was deleted or left blank), while the `full_neo4j_data` / `full_pg_data` volume persisted with the original password. The stack then came up with an unhealthy Neo4j (`Unauthorized`) or a Postgres that rejected every query — and the installer only warned **after** a failed launch.

Now, before generating any secret, the installer detects a data volume that exists while its password is unrecoverable (blank in `.env`, or the shipped `.env.example` placeholder) and stops with an explicit choice — restore the original password, or reset the volume with `down -v`. The abort happens before staging the temp `.env`, so nothing half-written is left behind. Both password-generating paths are guarded: `configure()` and `backfill_env_secrets()`. New helpers `orphan_db_volume()` and `assert_recoverable_db_passwords()` encapsulate the check.

## 2. Regenerate DB passwords on a full reset (the "cached password" report)

A "reset" only removed the Docker volumes; the DB password lived on in `.env`, and `resolve_secret` deliberately reuses it (to avoid rotating a live database). So after `--fresh-docker-reset` — or the resume "reset" action, or a manual `down -v` + reinstall — `NEO4J_PASSWORD` / `POSTGRES_PASSWORD` came back **unchanged, as if cached**. If the original credential was the problem, clearing Docker never fixed it.

When the data volumes have just been wiped, the passwords in a surviving `.env` are meaningless. A new `RESET_DB_SECRETS` flag — set by `fresh_docker_reset()` and the resume "reset" action — makes `configure()` drop and regenerate the DB passwords. Other secrets (MCP API key, Fernet key, JWT secret) are preserved (not tied to the wiped volumes; rotating the MCP key would gratuitously un-wire the agent). The blanking happens **before** the orphan-volume guard, so it composes safely: if `down -v` silently failed and the volume survived, the guard sees a blank password against a live volume and aborts instead of writing a guaranteed mismatch.

## Tests

Full installer suite green:
- `test_install.sh` 50/50 — includes the new full-reset regeneration case and the "reset failed to remove the volume → abort" safety case.
- `test_resume.sh` 61/61 — guard unit + integration cases, `fresh_docker_reset` sets the reset flag.
- `test_failgen.sh` 13/13, `test_connect_hermes.sh` 57/57, `test_connect_openclaw.sh` 38/38, `test_backup_file.sh` 5/5.

Also verified end-to-end against real `docker`: a fresh install over orphaned volumes aborts with guidance (no `.env`, non-zero exit); a full reset regenerates the DB passwords while preserving the MCP key.

## Docs

`install.md` troubleshooting: the "cannot be recovered" abort, and a new "reset didn't change the password" section explaining the two places the password lives and how to force new ones.

## Note

Rebased onto current `main` (after the OpenClaw integration merge, #270). Clean rebase, no conflicts.


toomij99:
## Review findings

Ran a multi-angle review over the diff (line-by-line scan, removed-behavior audit, cross-file trace, reuse/simplification/efficiency/altitude passes) and independently verified the candidates against the code. Six survived verification — all confirmed against the actual PR-head source. Please address before merge, ranked most-severe first:

1. **`resolve_volume_name()`'s suffix match isn't scoped to this Compose project** (`install.sh` ~L1462). `docker volume ls | grep -E "_${short}\$"` matches *any* volume ending in `_full_neo4j_data`/`_full_pg_data`, not just this checkout's. Since `docker-compose.yml` has no `name:` field, the project prefix defaults to the directory name — a second clone/fork/colleague checkout in a differently-named directory produces `other-dir_full_neo4j_data`, which this regex still matches. A genuinely fresh install (no `.env`, blank password) in a new directory can hard-abort with "a database data volume from a previous install already exists" even though no volume for *this* install exists. Suggest scoping the match to the actual Compose project name (`docker compose ... config --volumes` already gets used elsewhere in the file, or `docker compose ... ls -q`/project name lookup) rather than a bare suffix grep.

2. **The guard fails open on a transient `docker` failure, indistinguishable from "no volume exists"** (`install.sh` ~L1462). `docker volume ls ... 2>/dev/null | grep ... || true` swallows both a failed daemon call and a genuine no-match into the same empty result, and the `docker volume inspect` fallback has the same issue. `check_prereqs()` only checks Docker once, early in the run; if the daemon hiccups later when `assert_recoverable_db_passwords` actually runs (e.g. right after `docker builder prune` in `fresh_docker_reset`), the guard silently concludes "no conflict" and a fresh random password gets written onto a volume that still has the old one — reproducing the exact bug this PR exists to prevent. Suggest distinguishing "docker call failed" (abort loudly) from "no matching volume" (proceed) rather than collapsing both to empty.

3. **`RESET_DB_SECRETS` is a hand-threaded flag with no derivation from real volume state.** The new guard independently queries real Docker state to decide whether to *abort*, but whether to *regenerate* the DB passwords is decided by a separately-maintained boolean set only at the two current call sites (`fresh_docker_reset()`, `do_resume()`'s reset branch). A future reset-like path that forgets to set it won't crash — it'll silently reuse the stale password on a genuinely fresh volume, reintroducing the "reset did nothing" bug this PR targets. Notably, this PR's own `install.md` addition documents a manual reset recipe (`down -v` + hand-clearing `.env` via `sed`) that only works today because it happens to blank the password by hand, not because it goes through the flag. Consider deriving "should regenerate" from the same volume-existence check the guard already computes, rather than a manually-threaded flag.

4. **`orphan_db_volume()` reimplements `resolve_secret()`'s recoverable-password predicate instead of sharing it** (`install.sh` ~L1479-1486 vs ~L1213-1220). The two conditions are exact De Morgan negations of each other, computed independently. If the blank/placeholder rule ever changes, both call sites need updating in lockstep or the guard's notion of "recoverable" silently diverges from `resolve_secret`'s. Suggest factoring out a shared `is_recoverable_secret(key, prev)` helper.

5. **The "shipped placeholder" branch is dead code for both keys, and doesn't cover the real dangerous default** (`install.sh` ~L1483, `install.md`). `.env.example` has no `NEO4J_PASSWORD=` line at all and ships `POSTGRES_PASSWORD=` blank, so `example_val` always returns `""` for both keys — the placeholder check degenerates to the blank check. Meanwhile the actual reachable dangerous default, `metronix_dev` (from `docker-compose.yml`'s `${VAR:-metronix_dev}` fallback on a bare `docker compose up`), isn't checked at all — contradicting the doc-comment and install.md prose claiming the guard catches "the shipped placeholder."

Could you take a pass at resolving these before merge? 1 and 2 are the ones I'd consider blocking — both can let the exact failure mode this PR is meant to prevent (or a false-positive abort of a legitimate install) slip through. 3-5 are lower-severity but worth addressing for robustness/maintainability.

toomij99 (commented):
## Automated review

**CI status:** ✅ All checks passing (Analyze js/ts, Analyze python, lint, pytest, ruff format+lint).

**Lint/test:** Not re-run locally (bash installer scripts + existing test harness), CI already covers this comprehensively.

**Code review:**
This hardens `install.sh` against a real footgun — reusing a stale DB password against a data volume that fixed a different password on first start, which would previously "succeed" and then fail auth at container startup. Findings:

- `assert_recoverable_db_passwords` / `orphan_db_volume` guard logic is sound: it only flags a conflict when a volume exists AND the current password is blank or still the `.env.example` placeholder, and it's invoked before any secrets are written (both in `configure()` and `backfill_env_secrets`), so a partial/broken state is never staged.
- `RESET_DB_SECRETS` is threaded correctly through `fresh_docker_reset` and `do_resume`'s reset path, blanking the previous password lookups so `configure()` regenerates fresh ones instead of reusing stale values — and it does so *before* the guard runs, so a reset that failed to actually remove the volume (`down -v` silently no-op'ing) still gets caught by the guard rather than shipping a mismatched password.
- Test coverage is thorough: new cases (13–14, R20–R25) cover the happy path (reset regenerates + preserves other secrets), the failure path (volume survives reset → abort, no launch), the guard function directly (blank/placeholder/real password × volume presence), and `backfill_env_secrets`'s own call site.
- `docker() { return 1; }` stubs added to existing sandboxed tests are correct and necessary now that the guard shells out to `docker volume ls` — without them these tests would leak into a developer's real Docker state.
- No blocking issues found.

Clean — no blocking issues. Nothing further needed from my side on this one.

toomij99:
## Follow-up on the review findings

Konstantin's [716835d](https://github.com/mtrnix/metronix-memory/commit/716835d) addressed all 5 findings from the earlier review. While independently verifying it, I found the installer test suite was red on that commit (`test_resume.sh`: 55/66 passing) — not caught by CI, since `.github/workflows/installer.yml` only runs `shellcheck` + `bash -n`, never `tests/installer/*.sh`.

**Root cause:** the new `compose_project_name()` fell back unconditionally to `basename($PWD)` when `$COMPOSE_PROJECT_NAME` wasn't exported in the shell. Two consequences:
- 4 test cases (R19, R20, R21, R24) run from a throwaway `mktemp` dir, so `basename($PWD)` never matched their `metronix-memory_...` fixture volumes — the exact-match volume lookup silently missed, making those tests pass through the "no volume" path instead of exercising the scenario they were written for.
- More importantly for production: real `docker compose` also honors `COMPOSE_PROJECT_NAME` set in `.env`, not just the shell environment. A user who sets it there (a normal, documented way to do it) would get a **silently wrong** project name here, causing the guard to miss their real volume entirely — a false "no volume" that reintroduces the exact auth-mismatch bug this PR fixes, just via a different path.

Pushed [ac1f805](https://github.com/mtrnix/metronix-memory/commit/ac1f805): `compose_project_name()` now also checks `.env`'s `COMPOSE_PROJECT_NAME` before falling back to the directory name, and the 4 test cases are pinned to `COMPOSE_PROJECT_NAME=metronix-memory` to match their fixtures.

All 230 installer tests pass; shellcheck clean; verified live against real Docker that `.env`-based `COMPOSE_PROJECT_NAME` is now correctly picked up.

toomij99 (approved):
All 5 findings from the earlier review are resolved cleanly:

1. Volume matching is now scoped to this project via `compose_project_name()` (checks `$COMPOSE_PROJECT_NAME`, then `.env`'s value, then falls back to the directory name) with an exact match (`grep -Fx`) instead of a bare suffix grep — no more false-positive aborts from an unrelated checkout's volume.
2. `probe_db_volume()` now returns a distinct rc=2 for "Docker unreachable" vs. empty for "no volume", and `assert_recoverable_db_passwords` aborts loudly via `docker_probe_failed()` instead of silently proceeding — closes the fail-open gap.
3. `RESET_DB_SECRETS` is gone; `configure()` now derives keep-vs-regenerate directly from real volume state (`probe_db_volume`), so any future path that wipes volumes gets correct behavior automatically instead of depending on remembering to set a flag.
4. `is_recoverable_secret()` is now the single shared predicate used by both `resolve_secret()` and `orphan_db_volume()` — no more duplicated logic that could drift apart.
5. The placeholder-detection limitation for the DB keys (`.env.example` ships no real placeholder for them) is now explicitly documented in the code comment and the doc prose no longer overclaims it.

Verified: CI green (lint, ruff, both CodeQL analyze jobs, pytest-with-services), and the follow-up commit's own test-suite fix (`compose_project_name` honoring `.env`'s `COMPOSE_PROJECT_NAME`) closes a real gap that would have silently broken the guard for users who set that var in `.env`. Nice fix overall — approving.

toomij99:
@Sergey-Vasnev could you take a look and approve this PR? All review findings have been addressed and CI is green (lint, ruff, CodeQL analyze x2, pytest-with-services all passing). Branch protection requires an approval from someone other than the PR author or the last committer to merge — thanks!

Sergey-Vasnev (approved):

PR #318, opened 2026-07-04
docs: simplify Star History link URL in README

toomij99:
Simplifies the Star History anchor URL in the README (drops the empty `logscale=` param and reorders query params).

Note: this does not affect the intermittently-empty chart rendering, which is caused by upstream rate-limiting on star-history.com's shared API, not the README markup.


toomij99:
Two different tasks in one PR. Hermes memory is too early to implement

Issue #343, opened 2026-07-20
feat(memory): extract durable candidate memories from completed agent turns


toomij99:
## Problem / motivation

Metronix stores and retrieves agent memory well, but durable memory creation is mostly explicit. Agents must decide to call `metronix_memory_store`, while the Hermes scaffold can mirror raw turns into session memory. This makes long-term recall incomplete or noisy.

## Proposed solution

Add an opt-in, privacy-first conversation-compaction foundation.

- Retain raw user/assistant/tool events temporarily (7 days by default; configurable to 24h, 7d, 30d, or forever).
- Compact a session at explicit, session-end, inactivity, event-count, or token-budget triggers.
- Persist source-linked session ledgers containing people, preferences, decisions, commitments, corrections, unresolved threads, and follow-ups.
- Extract only durable facts, decisions, preferences, and explicit instructions.
- Write inferred records as `candidate`; write explicit "remember this" records as `active`.
- Include source hashes, extraction version, confidence, and timestamps.
- After raw-event expiry, retain only non-reversible provenance metadata—not raw text.
- Default durable output to `per_agent`; sharing is an explicit promotion.

## Acceptance criteria

- Feature is disabled by default.
- Extraction/compaction is idempotent per source event range.
- Candidates have evidence/provenance and appear in the existing review queue.
- Candidate records are typed, source-linked, and privacy-scoped; inferred candidates cannot affect agent behavior until reviewed or explicitly policy-approved.
- Explicit opt-in instructions can be persisted directly.
- The prompt assembler can inject a compact session ledger without injecting expired raw event text.
- Integration tests cover extraction, event expiry, deduplication, agent/workspace isolation, and feature-disabled behavior.

## Related issues

- #344 defines the automatic-write privacy and quality gate.
- #345 adds reviewed promotion of useful session memory.
- #346 adds provenance, corrections, and retrieval-quality feedback.
- #347 depends on this foundation for portable Agent Memory Bundle migration.
- #348 consumes durable compacted records for Time Machine history.

## Alternatives considered

Continue relying on routing prompts and client-side calls to `metronix_memory_store`. This remains supported but does not provide a consistent automatic path.

hegu-1:
Extracting durable candidate memories from completed turns is the right default: capture broadly, promote carefully.

I would make the compaction output explicitly candidate-level first:

- people/preferences/decisions/commitments/corrections as separate types
- source links to raw events
- confidence and privacy scope
[...]

PR #351, opened 2026-07-20
feat(memory): add conversation compaction foundation

toomij99:
## Summary

Implements the conversation-compaction foundation for general conversational agents (#343).

- Stores temporary conversation events in PostgreSQL with a 7-day default retention and 24h, 30d, and forever options.
- Adds fail-closed local filtering for secrets, credential-like values, and untrusted instruction content before temporary or durable persistence.
- Adds source-linked, hash-only session ledgers; durable candidate memories are per-agent by default and require explicit promotion for broader scope.
- Adds authenticated explicit session compaction, numeric-only ledger prompt injection, safe SSE capture after a framed terminal `[DONE]`, and opt-in expiry.
- Adds a durable claim/finalize/recovery lifecycle so explicit compaction processes each event batch once and history replay captures only the new exchange.

## Safety posture

Capture-triggered automatic compaction remains disabled pending a future cross-process claim-lease renewal design. Explicit compaction is available now; raw event content is never injected into prompts.

## Validation

- `POSTGRES_HOST=localhost POSTGRES_PORT=5433 uv run --extra dev pytest ...` — 61 passed
- Ruff check and format check passed
- Scoped mypy passed
- Alembic reports migration `031` as the single head

The focused API tests emit six existing `datetime.utcnow()` deprecation warnings.


toomij99 (commented):
Blocking cleanup note: please remove the five `.superpowers/sdd/task-*-report.md` files from this feature PR. They are internal execution/hardening reports rather than product documentation, and publishing them expands the public repository surface with implementation-process material unrelated to the conversation-compaction feature. Keep any user-facing lifecycle documentation under `docs/` instead.

PR #378, opened 2026-08-06
fix(auth): accept personal API keys for Hermes memory REST calls

toomij99:
## Summary
- Accept personal `mtk_…` API keys on protected REST routes as an alternative to JWTs, unblocking Hermes Metronix memory reads and writes.
- Resolve the key owner and preserve its active status, role, and workspace grants; unknown, revoked, and inactive-owner keys remain unauthorized.
- Add focused REST authentication coverage for valid, revoked, inactive-owner, and static OpenAI-compatible keys.
- Retain the documentation/demo assets and KB Admin teardown guidance already present on this branch.

## Motivation
Hermes could authenticate successfully with the Metronix provider configuration but its personal API key was rejected by `/api/v1/memory/*`, causing 401 errors during prefetch and turn sync.

## Design
JWT validation remains the first path. When it does not validate, the REST authentication dependency validates a personal API key and derives the same user principal used by protected routes.

## Validation
- `git diff origin/main...HEAD --check` passed.
- Added `tests/unit/test_personal_api_key_auth.py`.
- Local pytest was not run in this isolated worktree because its Python environment is absent; GitHub Actions has been re-triggered for the pushed commit.

## Scope
The primary runtime change is personal API-key support for protected REST endpoints. No changes were made to the Hermes plugin.


toomij99 (commented):
The new `docs/readme-preview.html` is a hand-maintained copy of `README.md`, but it is already out of sync with this same PR: it omits the new KB Admin profile teardown section added to `README.md`. It also has no generator or validation that would keep the copy current as the README changes. Please generate the preview from the README as part of a documented workflow (and verify it in CI), or remove the duplicate artifact and use a preview approach with a single source of truth.

PR #399, opened 2026-08-14
docs: document required FERNET_KEY for manual install

cycterna2222288888-ai:
## Summary

`FERNET_KEY` was missing from the manual-install docs as a required variable. Without it,
`metronix_source_create` (and any other connector-add path) fails as soon as it tries to
encrypt a credential. The MCP client's error `code` is a generic `INTERNAL_ERROR` either
way, but the `message` field is descriptive: `metronix_source_create: FERNET_KEY not
configured. Set the FERNET_KEY env var.`

`install.sh` already auto-generates `FERNET_KEY` on a scripted install, and `.env.example`
documents it in a comment — but the manual-install path in both `README.md` and `install.md`
only called out `METRONIX_MCP_API_KEY` as required, so anyone following the manual steps
would hit that error on their first `metronix_source_create` call.

## Changes

- **README.md** — added `FERNET_KEY` alongside `METRONIX_MCP_API_KEY` in the manual-install
  `.env` block (step 2), with the `cryptography.fernet` generation command and a note on the
  `INTERNAL_ERROR` failure mode.
- **install.md**
  - Updated the overview steps, intro paragraph, and scenario table in "3. Configure `.env`"
    to list `FERNET_KEY` as required alongside the MCP key.
  - Added a new **3b. Fernet key (required for sources)** subsection, matching the style of
    the existing 3a (MCP key) subsection: generation command, what it protects, and the
    warning against using a non-Fernet placeholder.
  - Renumbered the subsequent subsections (chat LLM 3b→3c, Neo4j 3c→3d, Ollama 3d→3e) and
    updated all four internal anchor references accordingly.

## Test plan

- [x] Verified the error-shape claim against source: `src/metronix/mcp/tools/_source_deps.py`
      raises a bare `ValueError("FERNET_KEY not configured...")` when `get_settings().fernet_key`
      is empty. `handle_tool_error` in `errors.py` doesn't map `ValueError` to a dedicated code,
      so the response `code` falls back to generic `INTERNAL_ERROR` — but `message` is built as
      `f"{tool_name}: {error_message}"`, preserving `str(exception)`, so the MCP client still
      receives the descriptive text (e.g. `metronix_source_create: FERNET_KEY not configured.
      Set the FERNET_KEY env var.`).
  - Reasoning: confirmed by reading the source directly, not just the docs — the current
    happy-path text implies MCP setup only needs `METRONIX_MCP_API_KEY`, but the actual
    failure occurs at first `metronix_source_create` call, not at boot, which is what made this
    gap easy to miss.
- [x] Confirmed `install.sh` already generates `FERNET_KEY` automatically (`gen_fernet`,
      lines ~1247, ~1427) and treats a blank value as an install-blocking issue (~line 1622) —
      so this PR only fixes the manual-install documentation gap, not the scripted path.
- [x] Grepped `install.md` after renumbering subsections to confirm no stale `#3b-optional-chat-llm`
      / `#3d-bundled-ollama-embeddings` anchors remain, and all `§3a`–`§3e` cross-references
      point at the correct section.
- [ ] No functional/runtime changes — docs only, so no automated test suite run.


README.md
```diff
@@ -169,8 +169,13 @@ a JWT obtained from the Metronix administrator instead. Embeddings for ingest co
 
 ```bash
 METRONIX_MCP_API_KEY=...   # generate: openssl rand -hex 32
+FERNET_KEY=...             # generate: python -c "from cryptography.fernet import Fernet; print(Fernet.generate_key().decode())"
 ```
 
+Both are required. `FERNET_KEY` encrypts connector credentials — without it, `metronix_source_create`
+fails for every source with a generic `INTERNAL_ERROR` (the underlying cause, `FERNET_KEY not
```
toomij99: This says the underlying `FERNET_KEY not configured` cause is not surfaced, but `handle_tool_error()` preserves `str(exception)` and returns it as `message=f"{tool_name}: {error_message}"`; `source_create.py` then returns that full `MCPError.to_dict()` payload. The response code is `INTERNAL_ERROR`, but the response message is descriptive (for example, `metronix_source_create: FERNET_KEY not configured. Set the FERNET_KEY env var.`). The same inaccurate claim appears in the new `install.md` subsection and the PR description. Please distinguish the generic error *code* from the descriptive error *message* and update all three places.

toomij99 (changes requested):
The documentation gap is worth fixing, but the stated MCP failure behavior is incorrect and should be corrected before merge.

README.md
cycterna2222288888-ai: You're right — confirmed by reading `handle_tool_error()` directly: it does `error_message = str(exception)` and returns `message=f"{tool_name}: {error_message}"`, so the exception text is preserved end-to-end. `source_create.py` catches the `ValueError` from `_source_deps.resolve()` and returns that full `MCPError.to_dict()` payload as-is. So the response `code` is a generic `INTERNAL_ERROR` (no dedicated code for this case), but `message` is descriptive: `metronix_source_create: FERNET_KEY not configured. Set the FERNET_KEY env var.`

Fixed the wording in all three places to distinguish the generic error *code* from the descriptive *message*: README.md, install.md, and the PR description.

toomij99 (approved):
The requested documentation correction is complete: the README, install guide, and PR description now accurately distinguish the generic `INTERNAL_ERROR` code from the descriptive error message. No remaining findings.

PR #400, opened 2026-08-16
docs: fix Admin Console compose profile (kb -> admin)

cycterna2222288888-ai:
Signed-off-by: Пётр Драчев <petrdrachev10@gmail.com>

The Web Console section documents `docker compose --profile kb up -d --build`, but `docker-compose.yml` declares the `frontend` service under the `admin` profile — there is no `kb` profile in the file at all.

Compose silently ignores an unknown profile name, so the command starts the base services and exits 0 while the console never comes up, with no error pointing at the cause.

Verified against `docker-compose.yml`: `frontend` → `admin`, `embedding-proxy` → `benchmarker`, `open-webui` → `openwebui`. Ran `--profile admin` locally and the console came up on :3000 as documented.

Three occurrences fixed: README.md and two in docs/readme-preview.html.


toomij99:
The `kb` → `admin` correction is valid, but `docs/readme-preview.html` is still inconsistent with the README it claims to represent. The preview uses “KB Admin” and `http://localhost:3000`, while the current README documents “Metronix Admin” served by Caddy at `https://localhost:3000`.

Please regenerate or fully synchronize the preview rather than updating only the Compose profile name.

cycterna2222288888-ai:
Fixed — replaced every remaining `KB Admin` / `KB Admin Console` / `KnowledgeBase-UI` mention with `Metronix Admin` / `Metronix Admin Console`, and every `http://localhost:3000` with `https://localhost:3000`, across all 8 occurrences in `docs/readme-preview.html` (the section heading, the intro paragraph, the compose command comment, the demo walkthrough, the Important URLs table, and the Documentation list). Checked each one against the current README wording (e.g. "Metronix Admin upload UI", not "Metronix Admin Console upload UI", to match README:372) so the preview is consistent with it rather than just find-and-replaced.

Pushed as a new commit (`2bf8d3c`, rebased onto latest `main`) — PTAL.

docs/readme-preview.html
```diff
@@ -119,7 +119,7 @@ <h3>3. Launch (first run builds images + pulls embedding model, ~10–15 min)</h
 <pre><code>docker compose up -d --build</code></pre>
 <h3>4. Verify</h3>
 <pre><code>curl http://localhost:8000/health</code></pre>
-<p>A healthy backend exposes the REST API, the OpenAI-compatible API at <code>:8000/v1</code>, and the MCP endpoint at <code>:8000/mcp</code> (default on the host: <code>http://localhost:8000/mcp</code> — the <code>metronix-full-api</code> container, path <code>/mcp</code>; from Docker network: <code>http://metronix-core:8000/mcp</code>). If you have installed the KnowledgeBase-UI (e.g. <a href="http://localhost:3000">http://localhost:3000</a>), log in with your Metronix credentials. Default credentials:</p>
+<p>A healthy backend exposes the REST API, the OpenAI-compatible API at <code>:8000/v1</code>, and the MCP endpoint at <code>:8000/mcp</code> (default on the host: <code>http://localhost:8000/mcp</code> — the <code>metronix-full-api</code> container, path <code>/mcp</code>; from Docker network: <code>http://metronix-core:8000/mcp</code>). If you have installed the Metronix Admin Console (e.g. <a href="https://localhost:3000">https://localhost:3000</a>), log in with your Metronix credentials. Default credentials:</p>
```
toomij99: The current README requires generating a unique administrator password in `.env` and explicitly warns against shared/default passwords, but this preview still directs users to the shared `admin@metronix.local` / `metronix` credentials immediately below. Because this paragraph is being updated in this PR, please regenerate or synchronize the complete section, including the unique-password command and seeded-email explanation.

toomij99 (changes requested):
The Compose profile, Admin Console naming, and HTTPS corrections are valid, and CI is green. One security-relevant preview inconsistency still needs correction before approval: the generated preview retains shared default credentials that the current README explicitly replaced with unique-password setup guidance.

Please synchronize that section with README.md, then request another review.

cycterna2222288888-ai:
Fixed — synchronized the credentials paragraph in `docs/readme-preview.html` (line 122, now expanded to 124) with the current README section it mirrors (README.md:208-214):

- Same lead-in about installing the Metronix Admin Console, now including the self-signed-cert-by-default note that was also missing.
- Replaced the shared `login: admin@metronix.local / pass: metronix` block with the actual current guidance: generate a unique `AUTH_PASSWORD" ($(openssl rand -base64 32))` before first launch, don't use a shared or default password.
- Added the seeded-administrator-email sentence as its own paragraph, matching the README.

Pushed as `f2f13fe`, rebased onto latest `main`.

One heads-up, not fixed in this commit since it's a different paragraph than the one flagged: the file's Quick Validation → Option A → Step A curl/PowerShell examples a bit further down (around what's now line 130) still hardcode `"password": "metronix"`, while README's own Step A (README.md:224-225) now says to use the seeded email + the unique `AUTH_PASSWORD` you just set — same underlying staleness, just in the next section. Happy to fold that in too if you'd rather have it in this PR instead of a follow-up.

toomij99:
Congratulations on your contribution, and thank you for helping improve Metronix Memory! 🎉

PR #402, opened 2026-08-17
feat(connectors): accept google_drive as an alias for gdrive

### LINKED ISSUE #298: Connector polish: support `google_drive` as an alias for `gdrive`
## Problem / motivation
The backend uses `gdrive` while parts of the frontend already refer to `google_drive`. That naming mismatch is easy to trip over in docs, UI, and future integrations.

## Proposed solution
Accept `google_drive` anywhere `gdrive` is currently supported, while preserving backward compatibility.

## Context
`src/metronix/connectors/registry.py` registers `gdrive`, while `frontend/src/components/sources/ConnectionDialog.tsx` already uses `google_drive` in icon/color maps.

## Acceptance criteria
- Allow `google_drive` anywhere `gdrive` is currently accepted for connection creation or schema lookup.
- Preserve backward compatibility for existing `gdrive` users.
- Update any affected docs or examples so both names are not presented inconsistently.
- Add unit coverage for alias registration or normalization behavior.

## Difficulty
Easy

## Alternatives considered
Standardize on `gdrive` everywhere instead. That is possible, but `google_drive` is more explicit and aligns better with contributor expectations.

cycterna2222288888-ai:
Closes #298.

## What

Accepts `google_drive` as an alias for the canonical `gdrive` connector type, in three commits:

1. **`feat(connectors)`** — the actual fix. `resolve_connector_type()` in `connectors/schemas.py` maps `google_drive` → `gdrive`, called at the two places a caller-supplied `connector_type` first enters the system: `metronix_source_create` (MCP) and `POST /connections` (REST, used by `ConnectionDialog.tsx`). Both resolve the alias *before* validation, so `CONNECTOR_SCHEMAS`, `ConnectorRegistry`, and the stored `connections.connector_type` column never see anything but `gdrive`.
2. **`fix(frontend)`** — a bug this surfaced along the way, see below.
3. **`docs(mcp)`** — mentions the alias in the `metronix_source_create` / `metronix_source_schemas` / `metronix_source_sync` tool descriptions, so an agent discovers it from the docstring instead of trial and error.

## Why normalize-at-the-edge, not a second registry key

Considered registering `google_drive` as its own key in `CONNECTOR_REGISTRY` / `CONNECTOR_SCHEMAS` (pointing at the same `GDriveConnector`) instead. Rejected it:

- `GET /connections/schemas` would then list Google Drive **twice**, and `ConnectionDialog.tsx`'s connector picker grid would render two identical "Google Drive" cards — i.e. it recreates exactly the naming confusion this issue is about, just one layer up.
- The DB would end up with `connector_type` split across both `gdrive` and `google_drive` for existing vs. new rows, permanently, since nothing would ever normalize it — two sources of truth forever instead of one.

Normalizing once at the two creation entry points keeps `gdrive` as the single canonical value everywhere downstream (schema lookup, `validate_config`, registry, DB). Existing `gdrive` connections and code paths are untouched — the alias is pure input sugar, it never appears in a stored row or an API response.

`connector_type` is immutable after creation (not present in `UpdateConnectionRequest`), so these two entry points are the complete surface — no update-path normalization needed.

## Bonus: a live icon bug this uncovered

While tracing where `connector_type` strings are consumed, found that `CONNECTOR_COLORS` / `CONNECTOR_ICONS` in both `ConnectionCard.tsx` and `ConnectionDialog.tsx` were keyed by `google_drive` — but `schema.type` / `connection.connector_type` from the backend has always been `gdrive`. Since the alias never flows back out of the API (see above), that key was simply never hit: Google Drive silently fell back to the generic 🔗 icon and default `#6366f1` color in both the connector picker and existing connection cards. Fixed by rekeying both maps to `gdrive`.

## Tests

- `test_connector_schemas_aliases.py` — resolver in isolation (maps alias, passes through canonical/unknown names, resolved alias is a registered schema).
- `test_mcp_source_tools.py` — new case asserting `metronix_source_create("google_drive", ...)` returns `connector_type: "gdrive"` **and** that the value handed to the fake store's `create_connection` is `"gdrive"`.
- `test_connections_create.py` (new file) — same DB-persistence assertion through the REST route (`POST /connections`), plus a control case proving the canonical `gdrive` still works unchanged.

## Test status

Per `CONTRIBUTING.md`, the unit suite needs live Postgres/Neo4j/Qdrant/Redis, so the reliable path is the PR's `Tests` workflow against ephemeral containers — leaving the full run to CI.

Locally: `make lint` clean on every commit. For `make test`, ran the touched/related files directly each time (`test_connector_schemas_aliases.py`, `test_mcp_source_tools.py`, `test_connections_create.py`) — 25/25 passing, no regressions. Frontend commit: `tsc --noEmit` + `eslint` clean, `vitest` 16/16.


toomij99 (commented):
**[P1] Normalize the alias during schema lookup as well**

`resolve_connector_type()` is applied by the REST and MCP creation paths, but `get_schema()` still calls `CONNECTOR_SCHEMAS.get(connector_type)` directly. As a result, `get_schema("google_drive")` returns `None`, so the PR does not yet satisfy issue #298’s acceptance criterion that `google_drive` work anywhere `gdrive` is accepted for connection creation or schema lookup.

Please normalize inside `get_schema()` (or otherwise make alias-aware schema lookup part of the public contract) and add a direct regression test for `get_schema("google_drive")`.

toomij99 (changes requested):
**[P1] Normalize the alias during schema lookup as well**

`resolve_connector_type()` is applied by the REST and MCP creation paths, but `get_schema()` still calls `CONNECTOR_SCHEMAS.get(connector_type)` directly. Therefore `get_schema("google_drive")` returns `None`, leaving issue #298’s schema-lookup acceptance criterion unmet.

Please make the public schema lookup alias-aware and add a direct regression test for `get_schema("google_drive")`.

cycterna2222288888-ai:
Fixed — `get_schema()` now resolves the alias itself:

```python
def get_schema(connector_type: str) -> ConnectorSchema | None:
    return CONNECTOR_SCHEMAS.get(resolve_connector_type(connector_type))
```

Went with your parenthetical suggestion (make it part of the public contract) rather than a local fix: `validate_config()`, `validate_config_for_update()`, `mask_secrets()`, and `merge_config()` — the other public schema-lookup functions in the same module — now all route through `get_schema()` instead of hitting `CONNECTOR_SCHEMAS` directly, so alias resolution lives in exactly one place instead of being copy-pasted five times.

Also audited every other direct `CONNECTOR_SCHEMAS` usage in the codebase per your 'other places' concern (`source_sync.py`, `source_schemas.py`, `source_create.py`, `migrate_env_connections.py`, `api/autosync.py`, `connections.py`'s list/update/test/sync handlers, `channels/manager.py`) — all of them key off either already-resolved caller input (the two entry points from the first commit) or an already-canonical DB-stored `connector_type` (immutable after creation, so it can never hold an alias), so nothing else needed changing. Details in the commit message.

New test: `get_schema("google_drive") is get_schema("gdrive")` (same object) + the unknown-type case.

Pushed as `52cb0a2`. `make lint` clean; ran the full set of tests touching this module and connection creation/update (schemas, connections create/update, gdrive/github connectors, MCP source tools, env-migration) — 100/100 passed, no regressions.

PR #403, opened 2026-08-18
docs: add JavaScript/TypeScript SDK quickstart guide

### LINKED ISSUE #297: Example: add a JavaScript/TypeScript SDK quickstart
## Problem / motivation
The docs already cover Python and Go SDK usage, but JavaScript and TypeScript are missing from the quickstart path. That leaves a very common integration audience without a first-class example.

## Proposed solution
Add a JavaScript/TypeScript guide that covers the simplest happy path for Metronix usage.

## Context
The docs index includes Python and Go guidance, but there is no equivalent quickstart for Node-based agent builders.

## Acceptance criteria
- Add a new guide under `docs/integrations/` or `docs/examples/` for JavaScript/TypeScript usage.
- Cover the simplest useful flow: health check, one chat call to `/v1`, and one MCP or REST example for memory/search.
- Match the structure used in `docs/integrations/sdk-python.md` and the integration template in `CONTRIBUTING.md`.
- Link the new guide from `docs/README.md`.

## Difficulty
Easy

## Alternatives considered
Point JavaScript users at the Python examples. That helps a little, but it still makes contributors translate basics before they can verify anything.

cycterna2222288888-ai:
Closes #297.

## What's added

- `docs/integrations/sdk-js.md` — a JavaScript/TypeScript quickstart guide covering all three surfaces (OpenAI-compatible `/v1`, REST `/api/v1`, MCP), with the reference blocks from `sdk-python.md` / `sdk-go.md` plus a new **Quickstart** section: three copy-paste `fetch()` examples (health check, one `/v1` chat completion, one REST memory/search call) and an MCP example using the official TypeScript SDK. Every request body, header, and endpoint path is copied line-for-line from `docs/API.md` (Health, OpenAI-compatible API, and Memory sections) — nothing invented.
- `docs/README.md` — linked under **Runtime Guides**, right after `sdk-go.md`.

## Why this guide is structurally wider than sdk-python.md / sdk-go.md

`sdk-python.md` and `sdk-go.md` are values-only reference sheets — Base URL / Model / Key blocks, Verify, Troubleshooting, Recommendation — with zero runnable code. That's fine for those two, but #297's acceptance criteria explicitly require covering "the simplest useful flow: health check, one chat call to `/v1`, and one MCP or REST example for memory/search" with actual examples. Matching the python/go structure *and* satisfying that criterion meant keeping their reference blocks as-is and adding a **Quickstart** section on top, rather than picking one or the other.

## Why `@modelcontextprotocol/client` v2, not `@modelcontextprotocol/sdk` v1

I initially recalled `@modelcontextprotocol/sdk` from memory. Per the task's own instruction to verify rather than assume, I `npm pack`'d both packages and inspected the shipped `.d.ts` / `.d.mts` directly instead of trusting memory:

- `@modelcontextprotocol/sdk` v1.30.0 — still published, not marked deprecated on npm, but its own successor's README says outright: *"This is v2 of the MCP TypeScript SDK. It replaces the monolithic `@modelcontextprotocol/sdk` package from v1."*
- `@modelcontextprotocol/client` v2.0.0 — the current stable line. Confirmed straight from `dist/index.d.mts`: `Client`, `StreamableHTTPClientTransport`, `.connect(transport)`, and `.callTool({name, arguments})` all exist with exactly that shape — the guide's MCP example uses the SDK's own embedded JSDoc usage snippets as the source of truth, not a guess.

Extra signal that v2 is the right pick here specifically: our own backend pins `mcp==2.0.0` in `uv.lock` — same protocol-generation line as the TS v2 client, not just "whatever's newest on npm."

## CORS troubleshooting entry

Added one troubleshooting bullet that doesn't exist in `sdk-python.md` / `sdk-go.md`: CORS errors when calling `/v1` or `/api/v1` directly from a browser, with a pointer to proxy through your own backend instead of shipping `METRONIX_OPENAI_COMPAT_KEY` / a personal API key to client-side code. Python and Go are always server-side, so this failure mode doesn't apply to them — JS is the odd one out among the three SDK guides in that it can run in a browser, so it gets a browser-specific warning the others don't need.

## Separate, out-of-scope finding

While verifying MCP tool names against source for the guide's MCP example, noticed `docs/MCP_API.md`'s "Agent Memory" section headers (`#### memory_store`, `#### memory_search`, etc.) are missing the `metronix_` prefix — stale relative to the actual tool names in `src/metronix/mcp/tools/memory_*.py` (confirmed: `metronix_memory_store`, `metronix_memory_search`, ...). This guide uses the verified, correctly-prefixed names throughout. Not fixing `MCP_API.md` itself here — out of scope for #297, flagging for a follow-up.

## Checks

`make lint`: clean (docs-only change, no Python touched).


docs/integrations/sdk-js.md
```diff
@@ -0,0 +1,164 @@
...
+);
+console.log(await res.json());
+```
+
+## Using MCP instead
+
+For tool-driven agent runtimes, use the official TypeScript SDK,
+[`@modelcontextprotocol/client`](https://www.npmjs.com/package/@modelcontextprotocol/client)
+(v2 — it replaces the older `@modelcontextprotocol/sdk` package):
+
```
toomij99: The guide correctly says that `AUTH_ENABLED=true` requires a user JWT and ignores `METRONIX_MCP_API_KEY`, but this copy-paste example always sends `METRONIX_MCP_API_KEY`. It will return 401 in hosted mode. Please use a neutral token variable or show separate local and hosted forms, and align the troubleshooting entry with both authentication modes.

docs/integrations/sdk-js.md
```diff
@@ -0,0 +1,164 @@
...
+http://localhost:8000/api/v1
+```
+
+## MCP values
+
+```text
+URL:            http://localhost:8000/mcp
+Authorization:  Bearer <METRONIX_MCP_API_KEY>
+X-Agent-Id:     <stable-js-agent-id>
+```
```
toomij99, line 57: Issue #297 requires this guide to follow the integration template in `CONTRIBUTING.md`, which mandates `Prerequisites` and `Setup` sections. Their absence also hides an important runtime constraint: `@modelcontextprotocol/client@2.0.0` declares Node.js >=20, while the guide only mentions Node 18+ near the fetch examples. Please add the required sections and state the Node requirement for the MCP example.

toomij99 (changes requested):
Thanks for adding the JavaScript/TypeScript quickstart. The endpoint shapes and SDK imports are otherwise well grounded, and CI is green. Two documentation correctness issues need to be addressed before approval: the MCP credential example fails in hosted auth mode, and the guide omits the required Prerequisites/Setup structure (including the SDK v2 Node.js runtime requirement).

cycterna2222288888-ai:
Fixed both:

**1. MCP credential example fails in hosted mode.** Switched to a neutral `METRONIX_MCP_TOKEN` env var instead of hardcoding `METRONIX_MCP_API_KEY`, with an inline comment on what to set it to in each mode:

```ts
// Local (AUTH_ENABLED=false): set METRONIX_MCP_TOKEN to your METRONIX_MCP_API_KEY.
// Hosted (AUTH_ENABLED=true): set it to a user JWT instead — the shared key is
// ignored there (see the auth-mode callout above).
Authorization: `Bearer ${process.env.METRONIX_MCP_TOKEN}`,
```

Also updated the `/mcp` Troubleshooting entry to cover both modes instead of only the local one.

**2. Missing Prerequisites/Setup + Node version.** Added both sections per the `CONTRIBUTING.md` template. Verified the Node requirement myself rather than taking it on faith — `npm view @modelcontextprotocol/client@2.0.0 engines` really does return `{ node: '>=20' }`. Prerequisites now states Node 18+ for the fetch-only Quickstart examples vs. **Node 20+** specifically for the MCP example, and Setup is where `npm install @modelcontextprotocol/client` now lives (the 'Using MCP instead' section points at it instead of duplicating the install command).

Pushed as `3b85d0a`. `make lint` clean (docs-only change).

cycterna2222288888-ai:
Review fixes pushed in 3b85d0a on 2026-08-19. CI is green (ruff + pytest). Could you take another look?

PR #416, opened 2026-08-22
feat: highlight expired sources in answers and retrieval (#181)

### LINKED ISSUE #181: Highlight expired sources in answers / retrieval
Flag sources that are expired/stale when they appear in answers or retrieval results, so users can judge freshness at a glance.

**Definition of Done:** expired sources are visibly marked in answer citations and retrieval output.

cycterna2222288888-ai:
## Summary

Implements #181 in three commits, one per layer:

1. `feat(storage): surface status/freshness_score on qdrant search results`
2. `feat(retrieval): mark expired sources in fragments and answer citations`
3. `feat(mcp): expose source status in metronix_search_fast`

## Main finding: the signal already existed, it just got dropped on the way out

The freshness worker (`ingestion/freshness/target_raw_document.py:sync_downstream_stores`)
already mirrors `raw_documents.status` onto **every chunk's Qdrant payload**
the moment the freshness pipeline (MTRNIX-313) runs. So `status` was never
missing from storage — it was sitting right there in `payload["status"]`
the whole time.

What it never survived was the trip back out to a user. Four places in a
row read the payload and silently dropped the field before it reached
anything visible:

1. `qdrant.py:_format_result` — builds the standard result dict from the
   payload, but only copies `title`/`type`/`url`/`date`/… — not `status`.
2. `search.py:_collect_frags` — builds per-fragment dicts for the LLM
   context, no `status` either.
3. `search.py:_append_sources` — builds the `📚 Sources:` block appended to
   the `/chat` answer, again no `status`.
4. `mcp/tools/search_fast.py:_hit_to_item` / `SearchFastItem` — no status
   field on the MCP response model at all.

Given that, **the whole fix is additive** — no new detection, no new
pipeline stage, nothing "discovered" freshly. Each commit just stops
dropping a field that was already computed and already correct.

## Layer 1 — storage (`core/models.py`, `storage/qdrant.py`)

- `is_expired_status()` in `core/models.py`: the single place that decides
  what "expired" means — `stale` / `superseded` / `archived`. (`candidate`,
  `conflicted`, `review_needed` are agent-memory workflow states, not a
  freshness judgment on KB content, so they're deliberately excluded.)
- `status` (defaulting to `"active"`) and `freshness_score` are now
  top-level fields on `_format_result`'s output, in both
  `QdrantVectorStore` and `AsyncQdrantVectorStore`. Purely additive — the
  full payload dict was already there, this just stops throwing away two
  of its fields.

## Layer 2 — retrieval (`retrieval/search.py`)

Two very different audiences read the same `is_expired` flag, so they get
two different treatments:

- **The LLM** (`_build_ctx`, fed into the prompt) gets a short, neutral
  `" (outdated)"` suffix on expired fragments — nothing more. No "don't
  trust this source," no "warning," no "unreliable." Models get skittish
  about alarmist framing in their own context and start hedging or
  refusing to answer instead of just noting the caveat. Covered by a test
  that asserts none of that phrasing ever leaks into the LLM-facing text.
- **The user** (`_append_sources`, the `📚 Sources:` block appended to the
  final answer *after* generation) gets the specific label —
  `⚠️ (outdated)` / `⚠️ (superseded)` / `⚠️ (archived)` — since this text
  never goes back into the model, there's no risk in being precise here.

`/chat`, `/chat/stream`, and both `openai_compat.py` paths (streaming and
non-streaming) all route through `_append_sources` via
`chat.py:extract_sources_section`, so all four surfaces get the citation
marking for free — no separate change needed per endpoint.

`return_trace`'s `fragments`/`source_results` inherit `status` for free
too, which covers the "retrieval output" half of the issue's DoD for
trace/benchmark consumers.

## Layer 3 — MCP (`mcp/tools/models.py`, `mcp/tools/search_fast.py`)

`SearchFastItem.status: str = "active"` — a raw status string, not an
`is_expired: bool`. Chose this deliberately: `metronix_search_fast` skips
answer generation entirely (that's the point of the fast path), so there's
no server-side citation footer to build the way `/chat` has one — the MCP
client (an agent) is the one deciding what to do with the result. An
agent reasoning over these results plausibly wants to treat `archived`
very differently from `stale` (e.g. "still cite it, but note it's the old
version" vs. "treat as informational only") — collapsing both into a
single boolean would throw away a distinction the client is in a much
better position to act on than we are. The docstring on `SearchFastItem`
enumerates all four values (`active`/`stale`/`superseded`/`archived`) and
what each means, plus the missing-status contract below.

## Missing status always means active — tested, not just commented

Every layer defaults a missing/absent `status` to `"active"` /
not-expired, and every layer has a dedicated test for it (not just a
comment): `test_missing_status_defaults_to_not_expired`,
`test_missing_status_defaults_to_active` (×2, storage + MCP),
`test_missing_status_defaults_to_active_and_not_expired`,
`test_missing_status_has_no_expiry_label`.

This matters because most chunks in an existing deployment predate the
freshness pipeline (MTRNIX-313) or come from a source it doesn't cover —
they have no `status` field in their payload at all. If "missing" were
ever treated as "expired," the entire pre-existing knowledge base would
light up with false-positive staleness warnings on day one of this
shipping. Absence of evidence is explicitly not evidence of staleness
here.

## Out of scope: freshness doesn't affect ranking yet

`compute_signal_score()` in `retrieval/scoring.py` already accepts a
`freshness` / `freshness_weight` pair for exactly this purpose, but
`freshness_weight` is hardcoded to `0.0` — so today, a `stale` chunk scores
and ranks identically to an `active` one; this PR only changes how it's
*displayed*, not where it lands in the result order. Wiring
`freshness_weight` up to actually downrank stale content is a separate,
follow-up-shaped piece of work and isn't part of this issue's DoD ("visibly
marked"), so it's deliberately left untouched here.

## Testing

- `tests/unit/core/test_is_expired_status.py` — `is_expired_status()` for
  every `LifecycleStatus` value, plus missing/unrecognized-status cases.
- `tests/unit/storage/test_qdrant_format_result_status.py` — both store
  classes surface `status`/`freshness_score`, missing status → `"active"`.
- `tests/unit/test_expired_source_marking.py` — `_collect_frags` /
  `_build_ctx` / `_append_sources`, including the "no alarmist phrasing in
  LLM context" assertion.
- `tests/unit/test_mcp_search_fast_tool.py` — status passthrough +
  missing-status default on `metronix_search_fast`.
- `ruff check` / `ruff format --check` clean on every changed file.
- Full regression sweep across `search.py`/`qdrant.py`/MCP consumers —
  green aside from two pre-existing failures unrelated to this change
  (`test_token_budget.py::...test_token_budget_applied_before_llm_call`,
  `test_name_aliases.py::...test_russian_nickname_triggers_assignee_search`),
  both requiring a live Ollama/Neo4j connection not available in this
  environment, and both failing identically on `main` before this branch
  (verified via `git stash`).

Closes #181.


toomij99 (changes requested):
Changes requested: expired-source citations are malformed in the OpenAI-compatible endpoint.

`_append_sources` appends the freshness suffix after the URL (for example, `https://example.test/doc ⚠️ (outdated)`). The OAI renderer parses every character after ` — ` as the URL, so it emits an invalid markdown link such as `[📄 Old doc](https://example.test/doc ⚠️ (outdated))`. I reproduced this against the current renderer.

Please keep the status label separate from the parsed URL (or update the parser/renderer accordingly), and add regression coverage for both non-streaming and streaming OpenAI-compatible chat responses with an expired source.

cycterna2222288888-ai:
Thanks for catching this — confirmed and fixed in b174dc9.

**What changed:** `_append_sources` was putting the `⚠️ (<label>)` suffix *after* the URL: `f"{icon} {title} — {url}{expiry_suffix}"`. Moved it to the title side of the separator instead: `f"{icon} {title}{expiry_suffix} — {url}"`.

**Why fix the producer, not the two parsers:** both `_parse_source` and `_sources_to_markdown` in `openai_compat.py` split each source line on the single `" — "` and take everything after it as the href verbatim. Moving the suffix to the title side keeps that split point exactly where it already was, so neither parser needed to change — and they stay decoupled from the exact `⚠️ (...)` format `_append_sources` happens to use today. Patching the parsers instead would have meant stripping a trailing status pattern out of the URL in two places, coupling them to that format.

**Visible effect:** the freshness marker now shows up in the link text itself instead of leaking into the href — `[Old doc ⚠️ (outdated)](https://example.test/doc)` — for both the inline `[$[title]$]` citation and the footer "Sources:" block.

**Tests added:**
- `test_expired_source_marking.py::TestAppendSourcesExpiryLabel::test_expiry_label_sits_before_the_dash_not_after_the_url` — locks the exact line shape (`"Old Doc ⚠️ (outdated) — https://x/1"`) so a regression back to appending after the URL fails at the unit level, not just via the OAI-layer tests.
- `test_openai_compat.py::TestChatCompletions::test_expired_source_renders_as_well_formed_link` (non-streaming) and `TestChatCompletionsStreaming::test_streaming_expired_source_renders_as_well_formed_link` (streaming) — both assert the well-formed link and the absence of the old `(outdated))` / `url ⚠️` artifacts end to end.

`make lint` and the affected suite (`test_expired_source_marking.py`, `test_openai_compat.py`, `test_diversify_results.py`, 69 tests) all pass.

**Known limitation, left out of scope:** if a title itself ever contains `" — "`, `_parse_source`'s `rfind` and `_sources_to_markdown`'s `partition` were already ambiguous about where the title ends and the URL begins — that's pre-existing behavior, unrelated to this bug, and not touched here.

PR #368, opened 2026-08-02
feat: enforce agent memory ownership

toomij99:
## Summary

- Enforce explicit server-side agent-memory ownership and delegated capability grants across MCP memory operations.
- Require a verified JWT or personal API-key principal; shared MCP API keys remain unable to access agent memory.
- Scope update, delete, review listing, and review resolution to workspace, agent, and record; persist content-free policy-decision audit metadata.
- Support personal API-key principals in both mounted and standalone HTTP MCP transports.

## Validation

- `ruff check src tests/unit/conftest.py tests/integration/mcp/test_agent_access_conformance.py`
- Focused unit, conformance, and Compose-backed integration gate: `69 passed`

## Notes

- Live integration tests used the running Compose service ports: Postgres 5433, Qdrant 6335/6336, and Ollama 11435.

PR #418, opened 2026-08-23
docs: add LangGraph integration guide and runnable memory example

### LINKED ISSUE #301: Integration: add a LangGraph guide and runnable memory example
## Problem / motivation
Metronix documents LangChain, but many teams now build directly on LangGraph. Without a dedicated guide, users have to infer how to combine stable agent IDs, workspace scoping, and memory calls in graph-based workflows.

## Proposed solution
Add a LangGraph integration guide plus a runnable example that demonstrates the recommended Metronix integration pattern.

## Context
The repo already documents LangChain and supports both MCP-native tools and an OpenAI-compatible endpoint, but it does not yet offer LangGraph-specific guidance.

## Acceptance criteria
- Add `docs/integrations/langgraph.md` following the project’s integration-guide template.
- Include one runnable example under `examples/` that demonstrates stable `agent_id`, workspace scoping, and at least one memory retrieve/store step.
- Document when to use MCP versus `/v1` for LangGraph flows.
- Link the guide from `README.md` and `docs/README.md`.

## Difficulty
Medium

## Alternatives considered
Tell users to adapt the LangChain guide. That is workable, but it misses LangGraph-specific state and orchestration patterns.

cycterna2222288888-ai:
## Summary

Adds a LangGraph integration guide plus a runnable example, per #301:

1. `docs(integrations): add LangGraph guide` — `docs/integrations/langgraph.md`
2. `docs(examples): add LangGraph retrieve->reason->store example` — `examples/langgraph_memory_example.py`
3. `docs: link LangGraph guide from README and docs index`

The guide covers when to use MCP (`langchain-mcp-adapters`) vs `/v1` for LangGraph
flows, with a stable `agent_id` and `workspace_id="MTRNIX"` throughout, as required by
the issue's acceptance criteria.

## The example was actually run, not just written

Verified end to end against a live `docker compose up` stack (Postgres, Qdrant, Neo4j,
Redis, SPLADE, Ollama, API — all healthy), in both branches of the graph's conditional
edge:

**Run 1 — empty memory for this agent:**
```
[retrieve] 0 existing memory record(s) for: 'Q3 deployment window for the LangGraph demo'
[reason] should_store=True
[store] stored memory id=af2b6f5981c74a47abffdcbdcd110789
```
`retrieve` finds nothing → `reason` decides to store → `store` writes the fact via
`metronix_memory_store` and gets back a real id.

**Run 2 — same query, memory now populated:**
```
[retrieve] 1 existing memory record(s) for: 'Q3 deployment window for the LangGraph demo'
[reason] should_store=False
```
`retrieve` now finds the record `metronix_memory_store` just wrote → `reason` correctly
skips the store step. This is the same graph, same code path, only the state of the
knowledge base changed — confirming the conditional edge actually branches on live MCP
data rather than always taking one path.

(Test memory records were deleted via `metronix_memory_delete` afterward so the
`langgraph-demo-agent` id the example uses starts clean for the next person who runs it.)

## Three things that surfaced only by running it — not just reading the docs

**1. `langchain-mcp-adapters` conflicts with Metronix's own `mcp` version if installed in the same environment.**
`langchain-mcp-adapters==0.3.2` pins an older MCP SDK than Metronix's server needs.
Installing it into the repo's own `.venv` silently downgraded `mcp` from `2.0.0` to
`1.29.0` — which then broke `metronix.mcp.server`'s own import
(`from mcp.server.mcpserver import MCPServer`, only present in `mcp>=2.0`) with
`ModuleNotFoundError`. Caught this because `make lint`/tests would have broken on the
same branch. Fix: install the LangGraph client packages in their own environment,
never the backend's. Documented as a callout in Prerequisites and Troubleshooting.

**2. The shared `METRONIX_MCP_API_KEY` cannot call any `metronix_memory_*` tool — and this is bigger than this guide.**
`metronix_memory_search`/`metronix_memory_store` (and every other `metronix_memory_*`
tool) reject the shared key with `AUTH_REQUIRED: unauthorized agent memory access`,
while `metronix_status` accepts the exact same key fine. Root cause: `require_agent_access`
(`mcp/tools/_agent_access.py`, added in #368/#370) requires a bound request principal,
and the shared-key auth path (`authenticate_http_request` in `mcp/auth.py`) only ever
authenticates the transport — it never binds one. A **personal API key**
(`POST /api/v1/users`, trusted as admin locally when `AUTH_ENABLED=false`) does bind a
principal and works.

This guide's Setup/Troubleshooting now document the personal-key requirement, but the
scope of the actual problem is wider than LangGraph: `docs/integrations/sdk-js.md`,
`sdk-python.md`, and `langchain.md` all currently document the shared key as sufficient
for `metronix_memory_store`/`metronix_memory_search` MCP calls, which — per this same
auth path — would fail identically for anyone following them today. That's a
docs/behavior mismatch pre-dating this PR and outside its scope to fix; **flagging it
here for a separate follow-up issue** rather than touching those three files as a
drive-by in a LangGraph PR.

**3. `langchain-mcp-adapters` tool calls resolve to raw MCP content blocks, not a parsed dict.**
`await tool.ainvoke(...)` returns
`[{"type": "text", "text": "<json-encoded string>"}]`, not the dict Metronix's tool
actually returned. A naive `.get("id")` / `.get("results")` on that list silently
no-ops instead of raising — which reads exactly like "zero results" or "no id", and cost
real debugging time before the shape was obvious from a raw `repr()`. The example wraps
this in a `_parse_tool_result()` helper (`json.loads(raw[0]["text"])`); the guide calls
it out explicitly in Setup so nobody else has to rediscover it from a `repr()` dump.

## Testing

- `ruff check` / `ruff format --check` clean on every changed file, after each commit.
- `examples/langgraph_memory_example.py` run twice against a live local stack — see
  the two transcripts above.
- Manual MCP calls (`metronix_status`, `metronix_memory_search`, `metronix_memory_store`,
  `metronix_memory_list`, `metronix_memory_delete`) used during debugging to isolate each
  of the three findings above before writing them into the guide.

Closes #301.


toomij99 (changes requested):
Requesting changes before merge:

1. Hosted setup: a normal user JWT or personal key is not sufficient for `metronix_memory_search` / `metronix_memory_store`. Non-admin principals need an active grant for the exact workspace and agent; otherwise the policy returns `AUTH_REQUIRED`. Please document using an admin credential or provisioning the necessary agent-access grant.

2. Verification: the example retrieves before the conditional store and has no post-store retrieval node. One run cannot print both a newly stored ID and a matching search result. Please change the guide to require two runs, or add a second retrieval after storing.

cycterna2222288888-ai:
Thanks for both — addressed in 65c14d5 and 65785c6.

**1. Hosted setup / non-admin grant.** Confirmed by code, not guessing: `AuthorizationEvaluator.authorize` (`src/metronix/auth/policy.py`) bypasses the grant check entirely for `role == "admin"` (`admin_override`); anyone else needs an active row in `agent_access_grants` for the exact `(workspace_id, agent_id)` pair or every `metronix_memory_*` call fails closed with `AUTH_REQUIRED`.

I went looking for how that grant actually gets provisioned so the guide could describe the real path instead of a guess — and there isn't one. `agent_access_grants` is written to in exactly one place in the whole codebase: migration `032_agent_access_grants`'s one-time backfill (`INSERT ... SELECT ... FROM agents WHERE created_by <> ''`), which only covered agents that already existed when that migration ran. `agents/service.py::create_agent` doesn't insert a grant for new agents, there's no admin route or MCP tool that creates/lists/revokes one, and even the project's own integration tests reach for a direct `INSERT INTO agent_access_grants` in test setup rather than any application code path.

So per your two options: documenting "provision the grant" honestly meant documenting that there's currently no way to do that outside a manual DB insert — which isn't something to put in an integration guide as the recommended path. I documented the "use admin credentials" side instead (Setup step 2 + the `AUTH_REQUIRED` Troubleshooting entry no longer implies a personal key/JWT is sufficient on its own), and filed the gap itself as #424 with the grep/code trail, since it's a real product gap beyond this one guide.

**2. Single-run verification.** Added a `verify` node after `store` (`store → verify → END`), built from a small factory (`make_retrieve_node`) that `retrieve` also uses, so the actual search-tool call/parsing isn't duplicated — it just writes into a separate `post_store_results` field instead of overwriting `search_results`. The final summary now prints found-before, `stored_id`, and found-after from one invocation, and the old "run it again" hint is gone.

`ruff check`/`ruff format` pass on both changed files (`examples/` isn't in `make lint`'s scope, ran it directly); no test suite covers this example script.

PR #423, opened 2026-08-27
test: cover invalid agent_id format in metronix_memory_get_context

### LINKED ISSUE #299: Tests: add focused coverage for `metronix_memory_get_context`
## Problem / motivation
`metronix_memory_get_context` is a high-value MCP entry point for agent runtimes, but several important branches are not explicitly pinned down in focused tests.

## Proposed solution
Add targeted unit tests for validation behavior, feature-flag-off behavior, and default parameter handling.

## Context
`src/metronix/mcp/tools/memory_context.py` contains branches for invalid agent IDs, blank queries, disabled memory injection, and `memory_top_k` fallback behavior.

## Acceptance criteria
- Add unit tests for invalid params and blank-input error responses.
- Add a test for the early-return path when `memory_injection_enabled` is disabled.
- Add a test proving `memory_top_k` falls back correctly when omitted.
- Keep the tests isolated and avoid requiring external services.

## Difficulty
Easy

## Alternatives considered
Rely on broader integration coverage. That helps with smoke testing, but it is less precise for catching regressions in tool-level validation.

cycterna2222288888-ai:
## What

Adds the one validation branch in `metronix_memory_get_context` that wasn't
covered by `tests/unit/test_memory_context_mcp.py`: a non-empty but malformed
`agent_id` (outside `A-Za-z0-9._-`), which hits the `is_valid_agent_id` check
and returns `INVALID_PARAMS` — distinct from the already-covered empty-`agent_id`
path.

Existing coverage before this PR: missing/empty `agent_id`, missing
`workspace_id`, missing/whitespace `query`, `memory_injection_enabled=False`,
`memory_top_k` fallback, ungranted workspace, and the happy path via
`AgentContextAssembler`. The only gap was the malformed-but-non-empty
`agent_id` branch (`src/metronix/mcp/tools/memory_context.py:48-56`).

## Why assert on the message text

The new test asserts both `"error" in result` and that the message contains
`"must be 1-64 chars"`. Presence of `error` alone isn't a strong enough check:
`agent_id="agent id"` is truthy, so if branch ordering ever changed (or a
future refactor merged the two `agent_id` checks), the test could keep
passing while silently exercising the empty-`agent_id` branch
(`INVALID_PARAMS: agent_id is required`) instead of the format-validation
branch it's meant to pin down. Checking the message text ties the assertion
to the specific branch.

## Scope

Test-only change — `src/metronix/mcp/tools/memory_context.py` is untouched.
One file changed: `tests/unit/test_memory_context_mcp.py`.

## Verification

- `make lint` — clean (ruff check + format)
- `pytest tests/unit/test_memory_context_mcp.py` — 9/9 passed

Closes #299


toomij99 (approved):
Reviewed: focused invalid agent_id validation coverage; no blocking findings.

PR #439, opened 2026-08-31
docs: add LlamaIndex integration guide and example (#304)

cycterna2222288888-ai:
## Summary

Adds a LlamaIndex integration guide plus a runnable example, per #304:

1. `docs(integrations): add LlamaIndex guide` — `docs/integrations/llamaindex.md`
2. `docs(examples): add LlamaIndex retriever + memory round-trip example` — `examples/llamaindex_metronix_example.py`
3. `docs: link LlamaIndex guide from README and docs index` (+ register it in `PUBLIC_MCP_MODE_DOCS`)

The guide positions Metronix as a retrieval and memory **backend** rather than a
redundant local index: a custom `BaseRetriever` over `metronix_search_fast`, the
`metronix_memory_*` tools as agent tools, and `/v1` as an OpenAI-compatible answer
surface. It includes a section contrasting the server-side hybrid pipeline
(dense + SPLADE + graph + RRF + reranker + freshness) with a default
`VectorStoreIndex`, and a Verify section that inspects grounded results and citations
— covering the issue's acceptance criteria.

Client packages are pinned and installed **in a separate environment** from the
backend; unlike `langchain-mcp-adapters`, `llama-index-tools-mcp` tracks the same
major `mcp` line Metronix pins (`>=2,<3`), so there is no forced-downgrade conflict.

## Related issue

Closes #304

## The example was actually run, not just written

Verified end to end against a live `docker compose up` stack (Postgres, Qdrant, Neo4j,
Redis, SPLADE, Ollama, API — all healthy), on a clean `metronix-core` image, with
`llama-index-core` 0.14.24 / `llama-index-tools-mcp` 0.5.1 /
`llama-index-llms-openai-like` 0.7.2 and `qwen2.5:3b` behind `/v1`:

```
[retrieve] 4 node(s) for 'How does metronix_search_fast differ from the full pipeline?'
  - KB-SEARCH-FAST         score=0.033  url=-
  - KB-RETRIEVAL-OVERVIEW  score=0.032  url=-
  - KB-AGENT-MEMORY        score=0.032  url=-
  - KB-URL-PROBE           score=0.031  url=-

[answer]
Metronix uses dense vectors, SPLADE sparse terms, and a knowledge graph ...
(+ a Sources footer and trace id from the /v1 pipeline)

[memory.store] id=201e5c990caf4e0eb349f1bf1d85fe52 deduped=False
[memory.search] 2 hit(s)
[...]

Issue #304, opened 2026-07-03
Integration: add a LlamaIndex guide for retrieval plus durable memory


toomij99:
## Problem / motivation
LlamaIndex users care about retrieval composition, but Metronix does not yet explain how its dense, sparse, graph, and freshness-aware layers should fit into a LlamaIndex workflow.

## Proposed solution
Add a LlamaIndex guide and runnable example that positions Metronix as a retrieval and memory backend rather than a redundant local index.

## Context
Metronix already offers hybrid retrieval and durable memory, but contributors currently have to reverse-engineer how to integrate those capabilities into LlamaIndex.

## Acceptance criteria
- Add `docs/integrations/llamaindex.md`.
- Include a runnable example that uses Metronix as the retrieval or memory layer in a LlamaIndex workflow.
- Explain how Metronix’s hybrid retrieval differs from a default local vector index.
- Document at least one verification step that inspects grounded results or returned citations.

## Difficulty
Medium

## Alternatives considered
Point users at generic REST or `/v1` examples. That works, but it does not answer the LlamaIndex-specific retrieval questions people actually have.

PR #440, opened 2026-08-31
fix(ingestion): preserve url passed via knowledge-store metadata

cycterna2222288888-ai:
## Summary

`store_document()` — shared by `POST /api/v1/knowledge/store` and the `metronix_store`
MCP tool — silently drops a `url` passed through the `metadata` dict.

It builds `Document(...)` with no `url=`, so `Document.url` is `""`. Then
`ingestion/pipeline.py` rebuilds each chunk's Qdrant payload as
`{**doc.metadata, "url": doc.url}` — the empty `doc.url` overwrites the `url` key that
came in via `metadata`. `storage/postgres.py` writes `raw_documents.url` from `doc.url`
the same way (the `metadata` jsonb keeps it, but the column and the retrievable Qdrant
payload do not).

**Net effect:** a KB document stored this way always retrieves (`metronix_search_fast`
and the full hybrid pipeline) with `url == ""`, so citations built from retrieval have
no link. Connector-synced sources are unaffected — connectors set `doc.url` directly.

## Fix

Lift `url` out of `metadata` into `Document(url=...)` in `store_document()`. One place;
both entry points benefit. `url` is kept in `metadata` too, so a caller that reads its
own metadata back still sees it.

## Related issue

Surfaced while writing the LlamaIndex guide — see #304 / #439. That PR documents the
gap and uses `doc_label` as the citation key; this PR closes it.

## Verification

Against a live `docker compose` stack, on a `metronix-core` image built from this
branch:

```
$ curl -s -X POST 'http://localhost:8000/api/v1/knowledge/store?workspace_id=MTRNIX' \
    -H "Authorization: Bearer $TOKEN" -H 'Content-Type: application/json' \
    -d '{"content":"...","doc_label":"KB-URLFIX-VERIFY","metadata":{"url":"https://example.com/verify-fix"}}'
{"success":true,"doc_label":"KB-URLFIX-VERIFY","chunks_stored":1}

# metronix_search_fast over MCP:
  KB-URLFIX-VERIFY   url='https://example.com/verify-fix'   <- was '' before this change
```

Before the fix the same flow returned `url=''` for the hit.

## Type of change

- [x] Bug fix

## Checklist

- [x] `ruff format` and `ruff check` pass locally (CI-pinned `ruff@0.15.11`)
- [x] Tests added — `tests/unit/ingestion/test_store.py`: `store_document` lifts `url`
      from metadata onto `Document.url` (persisted + ingested), and leaves it empty when
      no `url` is supplied. `pytest tests/unit/ingestion tests/unit/api/test_knowledge*` → 46 passed
- [x] Docs updated — behavior is covered by #439 (LlamaIndex guide)
- [x] Changes are workspace-scoped (no change to scoping; `store_document` already takes `workspace_id`)


toomij99 (changes requested):
The new Document.url assignment fixes a new-document ingest, but it does not persist a URL when an existing document is re-stored with the same content and doc_label. `PostgresStore.upsert_raw_documents()` takes the unchanged-content branch and updates only `fetched_at`, so `raw_documents.url` and metadata remain stale. The immediate Qdrant write has the URL, but a later reindex rebuilds the citation from the stale source-of-truth row and loses it again.

Please update persistence semantics so URL/metadata changes update the raw document and reset Qdrant sync when required. Add a regression test: store a stable doc label and content without a URL, then re-store the same label/content with metadata.url; assert the raw row has the URL and a subsequent reindex retains it.

cycterna2222288888-ai:
@toomij99 Fixed in 0c3f1ce.

`upsert_raw_documents`' unchanged-content branch now checks whether the fields that also land in the Qdrant chunk payload — `url`, `title`, `author`, `metadata` — drifted from the stored row. When they have, it updates the row and clears `qdrant_synced` / `qdrant_synced_at` so a reindex (or the sweeper) re-embeds from a correct source of truth. `graph_synced` is deliberately left alone: graph entities/relationships derive from `content`, which by definition hasn't moved, and an LLM re-extraction on every metadata tweak isn't worth it. The row still counts as `unchanged`, but its `source_id` comes back in `changed_source_ids` so the caller re-ingests it.

This sits at the persistence layer, so it also covers a connector that re-fetches a content-identical doc whose upstream title/url changed — not just the knowledge-store path.

Regression tests in `tests/unit/test_raw_documents.py` (real `upsert_raw_documents` SQL over the in-memory SQLite engine):
- `test_upsert_unchanged_content_refreshes_drifted_url` — store stable content+label with no url, mark it synced, re-store the same content+label with `metadata.url`; asserts the raw row now carries the url, `qdrant_synced` is cleared (graph left set), and `get_unsynced_documents(target="qdrant")` — i.e. what a reindex reads — returns the row with the url.
- `test_upsert_unchanged_content_and_sidecar_only_bumps_fetched_at` — identical re-fetch, nothing drifted: flags stay, not queued for re-ingest.

Two side fixes needed to get the affected module green: three `upsert` assertions predated `changed_source_ids` and asserted the old 3-key dict; and the SQLite test shim only stripped `::jsonb`, not `CAST(:x AS jsonb)`, whose NUMERIC affinity was silently coercing JSON objects to `0`.

Still red in that file and out of scope here: `TestGetUnsyncedDocuments::test_get_unsynced_graph` and `TestMarkDocumentsSynced::test_mark_synced_by_source` fail on `main` too — the SQLite schema in the test lacks the `graph_failed` column that `get_unsynced_documents(target="graph")` references (added in #258). Happy to fix in a follow-up.

`ruff check` / `ruff format --check` clean on `postgres.py` and the test file.

PR #448, opened 2026-09-03
fix(retrieval): wire freshness_weight into signal scoring (#417)

### LINKED ISSUE #417: freshness signal computed but never applied — freshness_weight hardcoded to 0.0
### What happened?

`compute_signal_score()` (`src/metronix/retrieval/scoring.py`) accepts
`freshness` and `freshness_weight` parameters — the freshness signal is
computed and the plumbing for a rank downgrade exists — but neither is
ever actually wired up from the search pipeline:

- `freshness_weight` defaults to `0.0` in the function signature, and
  `search.py`'s call site only spreads `**_scoring_weights` from
  `get_profile_weights()` (`query_classifier.py`'s `QUERY_PROFILE_WEIGHTS`),
  which has no `freshness_weight` key at all — so the default is never
  overridden, for any query profile.
- There *is* a `Settings.freshness_weight` field (`core/config.py:391`,
  env `METRONIX_FRESHNESS_WEIGHT`, described as "Scoring weight for the
  freshness signal. 0.0 = off.") that looks like the intended on/off
  switch — but `search.py` never reads it. Setting
  `METRONIX_FRESHNESS_WEIGHT` to a non-zero value in `.env` has **zero
  effect** anywhere in the codebase; it's a fully orphaned config knob.
- `freshness` (the per-candidate signal itself) is also never passed at
  the call site, so it silently falls back to the function's own default
  of `1.0` for every candidate regardless of actual freshness_score.

Meanwhile, `search.py:_build_freshness_filter`'s docstring explicitly says
STALE documents are deliberately *not* excluded from search because
"scoring handles the rank downgrade" — but no such downgrade happens in
practice. Net effect: a `stale` chunk and an `active` chunk with identical
channel scores/recency/balance get the exact same `signal_score` and rank
identically.

**Expected:** either wire `Settings.freshness_weight` through to the
`compute_signal_score()` call (with a non-zero default, and the per-chunk
`freshness_score` actually passed as `freshness=`), or correct the
`_build_freshness_filter` docstring so it no longer claims a downgrade
that isn't happening.

### Steps to reproduce

1. Open `src/metronix/retrieval/scoring.py` — `compute_signal_score()`
   signature: `freshness_weight: float = 0.0`.
2. Open `src/metronix/retrieval/search.py` — the `compute_signal_score(...)`
   call site only passes `channel_scores`, `recency`, `balance`,
   `active_channels`, and `**_scoring_weights`; no `freshness=` and no
   `freshness_weight=` override.
3. `grep -rn "freshness_weight" src/metronix --include="*.py"` — only 3
   hits outside scoring.py's own docstring/formula: the `Settings` field
   definition (`core/config.py`) and the two usages inside
   `compute_signal_score()` itself. No call site in `search.py` (or
   anywhere else) ever reads `Settings.freshness_weight` or passes
   `freshness_weight=`.
4. `grep -n "QUERY_PROFILE_WEIGHTS" -A15 src/metronix/retrieval/query_classifier.py`
   — confirm none of the per-profile weight presets (`execution`,
   `documentation`, etc.) include a `freshness_weight` key.
5. Result: setting `METRONIX_FRESHNESS_WEIGHT` in `.env` and re-running a
   query changes nothing — `stale`/`archived` chunks rank exactly where
   their dense/graph/metadata/recency/balance scores would otherwise put
   them, with no freshness-based downgrade.

### Relevant logs / output

```
$ grep -rn "freshness_weight" src/metronix --include="*.py" | grep -v test
src/metronix/core/config.py:391:    freshness_weight: float = Field(
src/metronix/retrieval/scoring.py:74:    freshness_weight: float = 0.0,
src/metronix/retrieval/scoring.py:85:    ``freshness_weight`` defaults to 0.0 — when unchanged, the formula is
src/metronix/retrieval/scoring.py:109:    raw = recency_weight * recency + balance_weight * balance + freshness_weight * freshness
src/metronix/retrieval/scoring.py:110:    weight_sum = recency_weight + balance_weight + freshness_weight
```

No call site anywhere passes `freshness_weight=` or `freshness=` into
`compute_signal_score()`.

### Version / commit

`496a7f6`

### Environment

Found while working on #181 (highlight expired sources in answers /
retrieval). Purely a code-inspection finding — reproducible by reading the
call graph, no runtime environment or live deployment needed.

---

Discussed on #416, where this was flagged and deliberately left out of
scope (that PR only changes how expired sources are *displayed*, not
where they rank).

cycterna2222288888-ai:
Closes #417.

## Problem

`compute_signal_score()` accepts `freshness`/`freshness_weight`, and the
formula is fully implemented (and unit-tested), but nothing ever passed
either at the call site in `search.py`. `Settings.freshness_weight`
(`METRONIX_FRESHNESS_WEIGHT`) was read nowhere — a fully orphaned config
knob. `QUERY_PROFILE_WEIGHTS` (per-query-profile weight presets) has no
`freshness_weight` key either, and since the classifier's `get_profile_weights()`
is called unconditionally (classifier on *or* off), there was no path through
which the weight could ever become non-zero.

Net effect: `_build_freshness_filter`'s docstring claims STALE chunks aren't
excluded from search "because scoring handles the rank downgrade" — but no
downgrade ever happened. A `stale` chunk and an `active` chunk with identical
dense/graph/metadata/recency/balance scores ranked identically.

## Fix

- `search.py`: read the per-chunk `freshness_score` already mirrored onto the
  Qdrant payload by the KB freshness producer (#416/MTRNIX-181), falling back
  to `1.0` (fully fresh) when absent — matching `compute_signal_score()`'s own
  "unknown == Phase A" default, not the `raw_documents` row default of `0.5`,
  so missing data is never treated as a downgrade. Pass it plus
  `Settings.freshness_weight` as explicit kwargs at the (only) scoring call
  site.
- `freshness_weight` is sourced **globally from Settings**, not per query
  profile — see "Why not touch `QUERY_PROFILE_WEIGHTS`" below.
- `config.py`: `Settings.freshness_weight` **stays `0.0` (off) by default**
  (see below); its description now documents the `min_signal_score`
  interaction. Operators opt in with `METRONIX_FRESHNESS_WEIGHT`.
- `_scoring_weights` (the dict spread from the profile preset into
  `compute_signal_score()`) now defensively excludes `"freshness_weight"` too
  (mirrors the existing `"blend_weight"` exclusion), so a future per-profile
  `freshness_weight` key can't collide with the explicit kwarg and raise
  `TypeError`. A new test (`test_no_profile_defines_freshness_weight`) fails
  loudly if that key ever appears in `QUERY_PROFILE_WEIGHTS`, rather than
  letting the exclusion silently drop it unnoticed.
- **RAG trace** (`trace.py` / `search.py`): the `merge_and_score` phase now
  serializes the freshness scoring inputs — `freshness_weight` is merged into
  the phase's `weights`, and each candidate carries the `freshness` value it
  was scored with (`None` when the chunk had no `signal_components` entry, not
  a silent `1.0`). Without this a traced `signal_score` could not be
  recomputed from the payload.

`compute_signal_score()`'s own `freshness_weight=0.0` function default is
untouched — it stays the Phase A-parity default relied on by
`test_scoring_with_freshness.py::test_freshness_weight_zero_is_identical_to_phase_a`
and other direct callers.

## Why the Settings default stays `0.0`

`freshness_weight` wires up a knob that did nothing before; it does **not**
turn the downgrade on. Even with the KB freshness producer
(`freshness_kb_enabled`) off — so every candidate falls back to the same
`freshness=1.0` — a non-zero `freshness_weight` adds the *same constant* to
both the numerator and the denominator of `compute_signal_score` for every
candidate. Ranking within a query is unchanged, but the **absolute
`signal_score` shifts**, so a deployment with `min_signal_score > 0` would see
candidates cross the confidence-filter threshold that didn't before. #417
asked to make the orphaned knob functional, not to change scoring for every
existing install — so it ships off, and enabling it is a deliberate,
calibration-aware choice (called out in the field description).

## Why `QUERY_PROFILE_WEIGHTS` is not touched

`QUERY_PROFILE_WEIGHTS` encodes query-*intent* tuning (how much to trust
dense vs. graph vs. metadata for an `execution` vs. `relationship` query),
grid-searched on a dataset that predates the freshness signal entirely.
Staleness is a document-*lifecycle* signal, orthogonal to query intent —
there's no product reason a `documentation` query should discount stale
sources differently than an `execution` query. Threading a per-profile
`freshness_weight` in would mean inventing untested values for 6 profiles
with no grid-search backing, right before a planned Phase 6 re-run of that
search. Sourcing it once, globally, from `Settings` (the same pattern already
used for `active_channels`) is deferred and won't need to be
unwound in that eventual re-run either.

## Tests

- `tests/unit/retrieval/test_scoring_with_freshness.py`,
  `tests/unit/test_scoring.py`,
  `tests/unit/retrieval/test_scoring_normalization.py` — unchanged, still
  green (formula-level, already covered freshness before this PR; this PR
  only wires the call site).
- `tests/unit/test_query_classifier.py` — new
  `test_no_profile_defines_freshness_weight` guard.
- `tests/unit/test_rag_trace.py` — new
  `test_build_merge_phase_serialises_the_freshness_inputs` (both the
  `freshness_weight` and the per-candidate `freshness`, including the `None`
  case); the existing merge-phase tests updated for the new kwarg.
- `tests/unit/test_search_rag_trace.py` — the `hybrid_search_and_answer`
  wiring test now asserts the `merge_and_score` phase carries
  `weights.freshness_weight` and per-candidate `freshness`.
- `ruff@0.15.11 check .` / `format --check .` clean.

## Test plan

- [ ] `make grid-search` (Phase 6, S-grid) — separately re-tune
      `QUERY_PROFILE_WEIGHTS` on an eval set that now includes freshness, if/when
      profile-specific freshness weighting turns out to be wanted.

## Review follow-ups (toomij99)

Addressed in `981c010`:
- **Point 1** — the `0.1` default changed absolute `signal_score` for every
  install and could move `min_signal_score` outcomes. Reverted to `0.0`; the
  knob is now functional but off, per #417's actual ask. Field description
  documents the threshold interaction.
- **Point 2** — the RAG trace no longer recorded `freshness_weight` or the
  per-candidate `freshness`, so a traced `signal_score` wasn't reproducible.
  `build_merge_phase` now serializes both; `search.py` passes
  `Settings.freshness_weight` through. Unit + wiring tests added.


toomij99 (changes requested):
Thanks — this successfully wires the scoring call, but please address these before merge:

1. The new default `freshness_weight=0.1` changes absolute `signal_score` values even when every candidate falls back to freshness `1.0`. That changes behavior for deployments using `min_signal_score > 0`: candidates can cross the confidence-filter threshold despite the PR description saying existing installs see no behavioral change. Either preserve the prior default, account for threshold calibration, or update the compatibility claim and add coverage.

2. RAG traces are no longer reproducible: scoring uses `freshness` and the global `freshness_weight`, but `weights=_scoring_weights` excludes the latter and `build_merge_phase()` does not serialize the former. Include both in the trace payload and add a changed-path trace test.

cycterna2222288888-ai:
@toomij99 Both points addressed in 981c010, and the PR description is updated to match.

**1 — default back to `0.0`.** You're right that `#417` asked for the knob to *work*, not to be *on*: even with the producer off and every candidate at `freshness=1.0`, `freshness_weight=0.1` adds the same constant to numerator and denominator, so ranking is stable but the absolute `signal_score` moves — and `min_signal_score > 0` deployments would see that as candidates crossing the confidence filter. `Settings.freshness_weight` now defaults to `0.0`; `compute_signal_score`'s own function default was already `0.0` and is untouched. The field description now spells out: *"Any non-zero value also shifts absolute signal_score magnitudes for every candidate, so re-check min_signal_score calibration when enabling it."* No behavioral change for existing installs, and no compat claim to keep honest.

**2 — trace reproducibility.** `build_merge_phase` now takes `freshness_weight` and:
- merges it into the phase's serialized `weights` (it's Settings-sourced, deliberately not in `_scoring_weights`/`QUERY_PROFILE_WEIGHTS`), so the `weights` block is the complete vector `compute_signal_score` actually saw;
- emits each candidate's `freshness` alongside `recency`/`balance` — `None` when the chunk has no `signal_components` entry, rather than a silent `1.0` that would misrepresent what was scored.

`search.py` passes `_s.freshness_weight` into the call.

Tests: `test_build_merge_phase_serialises_the_freshness_inputs` (unit — both values, plus the `None` path) and the `hybrid_search_and_answer` wiring test now asserts `merge_and_score` carries `weights.freshness_weight` and per-candidate `freshness`. `ruff@0.15.11 check .`/`format --check .` clean; trace + scoring + classifier suites green (163 + 34 passed across the touched areas).

Issue #449, opened 2026-09-03
docs(integrations): extend the AUTH_REQUIRED credential contract past prompt-1/prompt-2


cycterna2222288888-ai:
Follow-up split out of #433 while implementing it — not addressed there, deliberately out of scope for that PR.

## Background

#433 fixed the credential-contract wording and preflight ordering in exactly `docs/integrations/{codex,hermes}/prompt-{1,2}-*.md` (4 files) — the ones `test_mcp_docs.py`'s doc-content regression suite did not fully cover.

Two adjacent gaps were noticed but not pulled into #433:

## 1. Coverage gap in `test_mcp_docs.py`

`docs/integrations/codex.md`, `docs/integrations/hermes.md`, and `docs/integrations/hermes-agent.md` (the non-numbered parent guides) already carry accurate wording from #419's fix, but `test_mcp_docs.py`'s `PUBLIC_MCP_MODE_DOCS`/`DOCS` lists don't include the numbered `prompt-*.md` files added/touched by #433 either. Decide whether to:
- add the 4 prompt files (and these 3 parent docs, if not already listed) to the existing parametrized lists, or
- keep a dedicated ordering-contract test suite (added in #434) separate from the general JWT/deployment-key wording suite.

## 2. `prompt-3-migrate.md` / `prompt-4-rollback.md` not audited

#433's scope was explicitly limited to prompt-1 (install) and prompt-2 (make memory mandatory). Nobody has checked whether `docs/integrations/{codex,hermes}/prompt-3-migrate.md` or `prompt-4-rollback.md` have the same "write before verify" ordering problem, or whether they even call `metronix_memory_*` tools in a way where it would matter. Needs a read-through of both files before deciding if there's anything to fix.

## Acceptance criteria

- A decision (and follow-through) on whether `codex.md`/`hermes.md`/`hermes-agent.md` plus the prompt-1/prompt-2 files get added to `test_mcp_docs.py`'s coverage.
- `prompt-3-migrate.md` and `prompt-4-rollback.md` (both codex and hermes) read through and either confirmed clean or fixed for the same ordering issue #433 addressed.

Related: #432, #433, #434.

Issue #434, opened 2026-08-30
test(mcp): cover the Prompt 1 to Prompt 2 authorization contract


toomij99:
## Problem\n\nExisting unit coverage proves that a personal key can resolve a principal and that a missing principal is denied, but the real user path spans static prompt templates, MCP request authentication, agent-memory authorization, and the Prompt-2 mutation boundary. This allowed the Prompt 1 -> restart -> Prompt 2 failure to ship.\n\n## Scope\n\nAdd focused regression coverage for the documented Codex/Hermes onboarding contract.\n\n- Assert every durable-memory prompt requires an agent-memory-capable credential and does not present METRONIX_MCP_API_KEY as sufficient.\n- Exercise a shared-key MCP request against a memory tool and assert the fail-closed AUTH_REQUIRED result.\n- Exercise the equivalent call with an administrator personal key or JWT and assert that the memory-list preflight can succeed for the configured workspace and agent.\n- Assert Prompt 2's policy-edit instructions are conditioned on that preflight; do not run a real Codex or Hermes installation in CI.\n- Keep tests secret-free and deterministic; use fixture-issued credentials only.\n\n## Acceptance criteria\n\n- The old shared-key-only wording or an unconditional Prompt-2 policy mutation fails a focused test.\n- The test makes clear that #424 remains the required follow-up for non-admin grants, while admin_override is valid for bootstrap.\n- The focused test suite passes in the repository's documented Python test environment.\n\nRelates to #419 and the onboarding plan issue.

Issue #432, opened 2026-08-30
Epic: unblock durable-memory prompt onboarding with an authenticated principal


toomij99:
## Goal\n\nMake the durable-memory onboarding path usable end to end: Prompt 2 must never make Metronix the required durable-memory store until the MCP client has a principal that can use metronix_memory_*.\n\n## Root cause\n\nThe local prompt-1 templates configure the shared METRONIX_MCP_API_KEY. That key authenticates MCP transport but resolves no principal, so Prompt 2 fails its memory verification with AUTH_REQUIRED. A personal mtk_ key or a user JWT resolves a principal; an administrator is authorized by admin_override.\n\nRelated:\n- #419 documents the shared-key versus agent-memory mismatch across integration guides.\n- #424 tracks runtime provisioning for non-admin agent grants. It is not required for the first-run administrator path.\n\n## Work plan\n\n1. Update Codex and Hermes prompt pairs to ask for or provision an agent-memory-capable bearer credential before enabling mandatory memory.\n2. Keep secrets out of generated prompts, logs, and durable agent instructions. The raw personal key is shown once by the server and must be stored only in the MCP-client configuration.\n3. Add a Prompt-2 preflight that clearly stops before changing AGENTS.md or SOUL.md when the memory channel is unauthorized, with an actionable remediation.\n4. Add an end-to-end regression that proves shared-key denial remains fail-closed and an authenticated administrator can complete the same memory check.\n5. Align the core MCP/auth documentation with the resulting onboarding contract and close the affected parts of #419.\n\n## Acceptance criteria\n\n- A new local deployment with no users has documented, safe instructions to seed admin@metronix.local via AUTH_PASSWORD without deleting data.\n- Prompt 1 does not claim that METRONIX_MCP_API_KEY alone makes agent-memory tools available.\n- Prompt 2 performs no durable-memory-policy edit until metronix_memory_list succeeds for its declared workspace and agent.\n- The supported administrator flow works with a personal mtk_ key or a valid user JWT and needs no agent-access grant.\n- A focused test covers both denied shared-key and successful principal-bearing calls.

toomij99:
Implementation work items: #433 (prompt contract and safe preflight) and #434 (regression coverage). Related existing work: #419 (cross-guide credential documentation) and #424 (non-admin agent-grant provisioning).

PR #450, opened 2026-09-03
fix(integrations): require an agent-memory principal before Codex/Hermes Prompt 2 (#433)

### LINKED ISSUE #433: fix(integrations): require an agent-memory principal before Codex/Hermes Prompt 2
## Problem\n\nThe Codex and Hermes prompt-1 templates configure METRONIX_MCP_API_KEY in the MCP Authorization header, then tell the user to restart and run Prompt 2. The shared key only authenticates transport and cannot call metronix_memory_*. Prompt 2 consequently writes or attempts to write a mandatory durable-memory policy and then fails its own metronix_memory_list verification.\n\nThis is the concrete onboarding manifestation of #419.\n\n## Scope\n\n- Update docs/integrations/codex/prompt-1-install.md and prompt-2-memory.md.\n- Update docs/integrations/hermes/prompt-1-install.md and prompt-2-memory.md.\n- State the required credential contract accurately: a user JWT or personal mtk_ API key is required for agent-memory operations; METRONIX_MCP_API_KEY may still be used for transport-only/non-memory use cases.\n- Document the first-start administrator seed only for an empty users table, using AUTH_PASSWORD. Do not recommend deleting volumes or printing passwords.\n- Make Prompt 2 verify authorization before modifying AGENTS.md or SOUL.md, and give a concise recovery path when denied.\n- Keep agent registration and agent-id ownership user-controlled; the prompt must not silently create agents.\n\n## Acceptance criteria\n\n- Following Prompt 1 then Prompt 2 with only a shared key cannot leave the client with an enforced but unusable mandatory-memory rule.\n- Following the documented administrator personal-key/JWT flow makes Prompt 2 verification succeed without an explicit grant.\n- The generated configuration never embeds a raw secret in AGENTS.md or SOUL.md.\n- The text is consistent with docs/MCP_API.md and does not weaken fail-closed authorization.\n\nRelates to #419 and the onboarding plan issue.

cycterna2222288888-ai:
Closes #433. Part of #432.

## Problem

`codex/prompt-1-install.md` + `prompt-2-memory.md` and the `hermes/` equivalents let a user finish onboarding with only `METRONIX_MCP_API_KEY` configured, then prompt-2 wrote the mandatory `metronix-config` block into `AGENTS.md`/`SOUL.md` *before* checking whether the memory channel was actually authorized (step 3 = write, step 4 = verify). The shared key resolves no principal, so `metronix_memory_*` always rejects it with `AUTH_REQUIRED` — this let a client end up enforcing a rule it could never satisfy, exactly as #432 describes.

## Fix

- `prompt-1-install.md` (codex, hermes): added a credential-contract note — `METRONIX_MCP_API_KEY` authenticates transport only; `metronix_memory_*` needs a personal API key (`mtk_…`) or a user JWT. Gives the exact paths: local (`POST /api/v1/users`, no auth needed when `AUTH_ENABLED=false`) and hosted (`/api/v1/auth/login` then `POST /api/v1/users/{user_id}/api-keys`). No secrets printed; points at the README's existing `AUTH_PASSWORD` seed flow rather than duplicating it.
- `prompt-2-memory.md` (codex, hermes): moved the `metronix_memory_list` check to a new **step 3 (preflight)**, before the `AGENTS.md`/`SOUL.md` write (now step 4).

  **Resolved explicitly in the doc text** (this PR's one open question going in): the success criterion is *absence of `error.code == "AUTH_REQUIRED"`*, not a non-empty `records` list — an empty memory list is success, not denial. Spelled out verbatim so a prompt-following agent can't misread "no records yet" as "not authorized."

  On `AUTH_REQUIRED`: STOP, do not touch the file, and give a concrete recovery path (get a personal key/JWT per prompt 1's note, update the `Authorization` header in the already-registered MCP config, restart, re-run). Step 5 (verify) drops the now-redundant second `metronix_memory_list` call — the preflight already proved it.

## Manual verification on a live stack

Stack was down; brought it up (`docker compose up -d`, all healthy). Then, with `AUTH_ENABLED=false`:

```
$ curl -X POST /api/v1/users -d '{"email":"...", "password":"...", "role":"admin"}'
# -> 201, {"...", "api_key": "mtk_..."}
```

Called `metronix_memory_list(workspace_id="MTRNIX", agent_id="...", limit=1)` over the real `/mcp` endpoint (MCP Python client), exactly as the new step 3 instructs:

- **Shared `METRONIX_MCP_API_KEY`:** `{"error": {"code": "AUTH_REQUIRED", "message": "metronix_memory_list: unauthorized agent memory access", ...}}` — a prompt-following agent stops here, before step 4.
- **Personal `mtk_…` key:** `{"records": [], "count": 0, "total": 0, ...}` — no `error` field, empty list. Confirms the "empty ≠ denied" criterion in the doc text is correct, not just asserted.

Test admin user created for the check was deleted afterward (`DELETE /api/v1/users/{id}`).

## Explicitly out of scope (filed as #449)

- Extending `test_mcp_docs.py`'s doc-content regression coverage to these 4 files plus `codex.md`/`hermes.md`/`hermes-agent.md`.
- Auditing `prompt-3-migrate.md` / `prompt-4-rollback.md` for the same write-before-verify ordering issue.

Regression coverage for the ordering/wording contract itself is #434, tracked as a follow-up PR stacked on this branch (diff will show only the new test files once this merges).

## Validation

- `make lint` — no Python touched, `ruff check`/`format --check` clean regardless.
- `tests/unit/test_mcp_docs.py` — 36 passed (these 4 files aren't in its parametrized doc list yet; that gap is #449).
- Manual live-stack check above.

## Test plan

- [ ] #434 adds regression coverage for the ordering + AUTH_REQUIRED/empty-list contract asserted here.

## Review follow-ups (toomij99)

Three review rounds found the preflight criterion itself was still fail-open, plus four related gaps. Addressed in `d80eec4`:

- **Fail-closed preflight** — success is now "response parsed successfully AND no `error` field at all," not "no error, or a non-AUTH_REQUIRED error." `WORKSPACE_NOT_FOUND` (an ungranted workspace — resolved, and deniable, before agent-access is even checked), a bare HTTP 401 with no `error` field, and any other error code now all STOP the agent before the file edit, same as `AUTH_REQUIRED` always did.
- **Re-run path** — the idempotent shortcut ("block already has the mandatory wording") used to skip straight to Verify, never re-checking authorization. It now still runs the preflight every time; it only skips the file edit.
- **Report line accuracy** — "memory channel reachable" is now attributed to the preflight step (which actually calls `metronix_memory_list`), not to Verify (which only calls `metronix_status`).
- **claude-code + root `prompts.md`** — both had the same write-before-verify shape (claude-code) or no preflight at all (`prompts.md`'s generic template). Fixed in this PR rather than deferred to #449, since #449 is test-coverage extension, not a bug fix.
- **`raw_key` vs `api_key`** — the admin-issued-key recovery path in `prompt-1-install.md` now names the correct response field.

**Not addressed here**, per the review's own non-blocking/tracked framing:
- The preflight requests `read` capability while the rule it installs mandates `write`-only usage, so a read-only grant still passes preflight and the first `metronix_memory_store` call is denied. This is a capability-check gap, not a fail-open/fail-closed one — needs its own fix (e.g. probing write capability directly), not just a wording change.
- Hermes' `prompt-1-install.md` OPTIONAL block still invites memory calls before prompt 2 runs, and the default `AUTH_ENABLED=false` path installs a shared key that's guaranteed to fail the preflight. Both are **#432**, already tracked there.


toomij99 (commented):
## Review: not ready to merge as-is

Great instinct on the fix, and the manual verification section is genuinely useful — but the new preflight has several fail-open gaps that let the exact #433 failure mode (mandatory rule written over an unauthorized/broken memory channel) recur through paths other than the one this PR closes. Details below, worst first.

### Blocking

**1. Preflight success criterion is fail-open** (`codex/prompt-2-memory.md:52`, `hermes/prompt-2-memory.md:54`)
`"the response has no error field, or its error.code is anything other than AUTH_REQUIRED"` passes on more than just success:
- `metronix_memory_list` resolves the workspace *before* checking agent access (`memory_list.py`), so a principal without a grant on `DEFAULT_WORKSPACE_ID` gets `WORKSPACE_NOT_FOUND`, not `AUTH_REQUIRED` — and that's exactly the shape a personal key/JWT scoped to the wrong workspace produces (the credential this PR's own recovery path tells the user to obtain).
- In hosted mode, an auth failure at the transport layer returns HTTP 401 with a `{"detail": ...}` body — no `error` field at all — which also satisfies "no error field = success."
- The preflight only requests **read** capability, but the rule it installs mandates **write**-only usage; a read-only grant sails through and then the first `metronix_memory_store` call is denied.

In all three cases, the agent proceeds to write the mandatory "never fall back locally" rule over a channel that doesn't actually work.

Suggested fix: invert to fail-closed — proceed only when there is a clean response with **no** `error` field; treat any error, any non-2xx/malformed response, or missing tool as a STOP, not just `AUTH_REQUIRED`.

**2. The idempotent re-run path skips the preflight entirely** (`codex/prompt-2-memory.md:4`)
When the mandatory block already exists, the header sends the agent straight to step 5, which only runs `metronix_status` (no `agent_id`, doesn't touch memory auth) — `metronix_memory_list` never runs. A credential revoked/rotated to broken since the last successful run goes undetected on re-run.

**3. Report line asserts a check that no longer exists** (`codex/prompt-2-memory.md:94`, `hermes/prompt-2-memory.md:100`)
Step 5 lost its `metronix_memory_list` call, but the Report format still requires printing "Verify: status ok, **memory channel reachable**." `metronix_status` can't observe memory auth at all, so this is now an unearned claim in every run — including the failure scenarios above, where it silently reports success.

**4. Two other runtimes sharing this bug were left unpatched**
- `docs/integrations/claude-code/prompt-2-memory.md:47` still writes the mandatory CLAUDE.md block before its step-4 verify — the pre-#433 ordering, untouched.
- Repo-root `prompts.md:166` (the generic template for Cursor, Claude Desktop, opencode, n8n, LibreChat, etc., and already in `test_mcp_docs.py`'s doc-contract list) has the same shape.

If the goal is "close #433," it's only closed for 2 of the runtimes that share the bug. Worth folding into this PR or a same-priority immediate follow-up rather than deferring alongside #449 (which is about test-coverage extension, not this).

### Worth fixing, non-blocking

**5. Hermes' unchanged OPTIONAL block still invites doomed memory calls** (`hermes/prompt-1-install.md:78`)
Prompt 1 still writes "You MAY use the metronix_* tools — knowledge search / RAG and memory... OPTIONAL" into the live SOUL.md using only the shared key. In the window between the mandated restart and running prompt 2, any interaction that triggers `metronix_memory_*` hits `AUTH_REQUIRED` with zero explanation in the loaded persona file (the new guidance lives only in prompt 2's text).

**6. Wrong response field name** (`codex/prompt-1-install.md:13`, same in hermes)
The note says the admin-issued key comes back as `api_key`; `POST /api/v1/users/{user_id}/api-keys` actually returns `raw_key`. An agent following the hosted recovery path and parsing for `api_key` finds nothing.

**7. Default local path guarantees a stop-and-recover round trip**
`POST /api/v1/users` needs no auth when `AUTH_ENABLED=false`, and prompt 1 already has shell access and already writes the Authorization header — but installs the doomed shared key anyway ("you do not need to fix it now"), guaranteeing every default-path user hits the preflight STOP and a manual mint-key/hand-edit-config/restart/re-run cycle. Real cost, but this is tracked under the still-open #432, so a note in the PR description pointing at that would be enough rather than a blocker here.

---

Recommend: fix #1–#3 (the preflight logic itself) before merge, since they undermine the PR's stated purpose in realistic scenarios (including the hosted/JWT path this PR recommends). #4 should at minimum get a tracking issue with the same urgency as this one, since it's the identical bug, still live.

cycterna2222288888-ai:
@toomij99 pushed `d80eec4` addressing this review:

**#1 Preflight success criterion (fail-open → fail-closed)**
Flipped the criterion in `codex/prompt-2-memory.md`, `hermes/prompt-2-memory.md`, and `claude-code/prompt-2-memory.md`: the agent now proceeds only when the response was parsed successfully **and** carries no `error` field at all. An empty `records` list still counts as success. Every other outcome — `WORKSPACE_NOT_FOUND`, `INVALID_PARAMS`, `INTERNAL_ERROR`, any other `error.code`, a non-2xx response, an unparseable body (the bare-401 case), or the tool not being available — is now an explicit STOP with no file edit, listed by name in the doc text. `AUTH_REQUIRED` keeps its specific recovery-path guidance on top of that.

Left out of this pass, on purpose: bullet 3 (preflight only requests read capability, rule mandates write-only) is a capability-check gap, not a fail-open/fail-closed one — flagging it but not fixing it here.

**#2 Idempotent re-run skipping preflight**
Fixed in codex/hermes: the re-run shortcut ("block already has the mandatory wording") now skips only step 4 (the file edit), not step 3 (the preflight). The header text says so explicitly, and step 3 itself says to run "every time this prompt runs."

**#3 Report line asserting a check that doesn't exist**
Re-attributed: "memory channel reachable" is now reported against step 3 (the preflight, which actually calls `metronix_memory_list`), separately from step 5's `metronix_status` result — instead of implying the Verify step proves it.

**#4 claude-code and root prompts.md left unpatched**
Both fixed in this PR:
- `docs/integrations/claude-code/prompt-2-memory.md` had the pre-#433 bug (write CLAUDE.md, verify after) — restructured to insert a preflight step 3 before the write (now step 4), matching codex/hermes.
- Root `prompts.md`'s generic Prompt 2 template (Cursor, Claude Desktop, opencode, n8n, LibreChat, …) had no preflight gate at all — added the same fail-closed step before the routing-rule edit.

**#6 Wrong response field name**
Clarified in both `prompt-1-install.md` files: the admin-issued key from `POST /api/v1/users/{user_id}/api-keys` comes back as `raw_key`; `api_key` only applies to the separate `POST /api/v1/users` self-serve path described in the same note. Verified against `src/metronix/api/routes/users.py`.

**#5 (Hermes OPTIONAL block) and #7 (default local path)**
Not fixed here — same non-blocking status as your review gave them. I'll add PR-description notes pointing at #432 for both, as you suggested, rather than rolling them into this PR.

**Verification:** `docs/integrations/{codex,hermes}/prompt-1-install.md`, `prompt-2-memory.md`, `docs/integrations/claude-code/prompt-2-memory.md`, and `prompts.md` — 36/36 existing tests in `tests/unit/test_mcp_docs.py` still pass unchanged (no doc-contract test currently covers claude-code/prompt-2-memory.md or prompts.md's Prompt 2 section directly — extending that is #449, not this PR). No Python source changed, so `ruff`/`mypy` are no-ops here.

toomij99 (changes requested):
The preflight only proves read access (`metronix_memory_list`), while the mandatory routing rule requires storing durable memory. A read-only grant therefore passes preflight, installs the no-local-fallback rule, and then fails on the first `metronix_memory_store` call, which requires write access. Please add a non-mutating write-capability probe and make the preflight require both read and write access before editing the routing rule.

cycterna2222288888-ai:
@toomij99 Addressed in 8ca76b8.

**There is a call that hits the write-authorization check before it hits any write:** `metronix_memory_delete`. It runs `require_agent_access(ws_id, agent_id, "write")` — the *same* gate `metronix_memory_store` uses — and only *then* touches storage; a `record_id` that matches nothing returns `DOCUMENT_NOT_FOUND` with zero rows read, written, or deleted (`memory_records.id` is free-form `TEXT`, and `MemoryService.delete` early-returns on `rowcount == 0` before any Qdrant/Neo4j/bus work).

So the new preflight is two checks, and step 4 (the file edit) is reached only if **both** pass:

- **3a — read:** `metronix_memory_list(...)`. Pass = parsed response, no `error` field (empty `records` still a pass).
- **3b — write, non-mutating:** `metronix_memory_delete(..., record_id="metronix-preflight-probe-<random hex>")`. The prefix + random suffix cannot collide with a real id (those are bare 32-hex). Pass = `error.code == "DOCUMENT_NOT_FOUND"` exactly — the write gate was cleared and the probe id simply matched nothing. `AUTH_REQUIRED` (incl. a read-only principal), any other code, a non-2xx/unparseable body, the tool missing, or — unexpectedly — a reported successful delete all STOP before the file edit.

`AUTH_REQUIRED` keeps its dedicated recovery path; the "don't special-case one code as the only failure" fail-closed framing from `d80eec4` is retained and now covers both sub-checks.

Applied to `codex` / `hermes` / `claude-code` `prompt-2-memory.md` and the generic `prompts.md` (steps 2a/2b there). Report-format lines updated. New `tests/unit/test_prompt2_memory_preflight.py` (12 cases) asserts both checks are present in all four files, the write probe is ordered before the `--- metronix-config ---` block, and the fail-closed + recovery wording survives.

`ruff@0.15.11 check .` / `format --check .` clean; `pytest tests/unit/test_mcp_docs.py tests/unit/test_prompt2_memory_preflight.py tests/unit/test_mcp_memory_tools.py tests/unit/mcp/` — 110 passed.

The read⊇write hierarchy note: a WRITE grant covers READ in `policy._covers` (`levels[granted] >= levels[requested]`), so 3b alone technically implies 3a — I kept 3a anyway for the clearer diagnostic when read works but write doesn't, and because it's the gentler first probe.

PR #451, opened 2026-09-03
test(mcp): cover the Prompt 1 to Prompt 2 authorization contract (#434)

### LINKED ISSUE #434: test(mcp): cover the Prompt 1 to Prompt 2 authorization contract

cycterna2222288888-ai:
Closes #434. Part of #432. Stacked on #450 (#433) — this branch includes that commit too, so the diff here will shrink to just the two new test files once #450 merges.

## Problem

Existing coverage (`test_mcp_agent_memory_auth.py`) proves the transport layer only: `authenticate_http_request` resolves `None` for a shared key and a real `MCPPrincipal` for a personal key. Nothing exercised:
- a real memory tool end to end (specifically `metronix_memory_list`, the exact call #450's new prompt-2 step 3 makes),
- the real `AuthorizationEvaluator`'s `admin_override` path (existing conformance tests use owner/delegate grants, not `role="admin"`),
- the prompt markdown's own ordering/wording contract that #450 just introduced.

## What this adds

**`tests/unit/mcp/tools/test_memory_list_preflight_contract.py`**
- `test_shared_key_preflight_is_denied_before_store_access` — no bound principal (shared-key equivalent) → `metronix_memory_list` → `AUTH_REQUIRED`, memory service never queried.
- `test_admin_principal_preflight_succeeds_via_admin_override` — admin principal (`workspace_ids=("*",)`, matching what `mcp/auth.py` actually derives for a fresh admin) through the **real** `AuthorizationEvaluator` (not a stub) → clean response, `records: []`, no `error` key. Proves `admin_override` end to end and that an empty list is not an error — the exact thing #450's doc text asserts.

**`tests/unit/test_prompt_onboarding_contract.py`** (same pattern as the existing `test_mcp_docs.py`)
- prompt-1 states the credential contract, never prints a password.
- prompt-2's preflight-call text precedes the policy-file-write text — **verified this actually catches a regression** before committing, by running the same ordering assertion against a synthetic "old buggy order" string; it failed as expected.
- the "empty ≠ denial" wording and the STOP/no-edit denial path are both present verbatim, so either regressing silently breaks CI.

Deliberately: no real Codex/Hermes installation is run; scope stays the same 4 files #450 touched (extending to `codex.md`/`hermes.md`/`hermes-agent.md` and auditing prompt-3/prompt-4 is #449, not here).

## Validation

- `ruff check` / `ruff format --check` on both new files — clean.
- `tests/unit/test_prompt_onboarding_contract.py` + `test_memory_list_preflight_contract.py` + `test_agent_access_guard.py` + `test_mcp_docs.py` + `test_mcp_agent_memory_auth.py` + `tests/integration/mcp/test_agent_access_conformance.py` together — 60 passed.


toomij99 (commented):
The preflight currently fails open for every error except `AUTH_REQUIRED`.

`metronix_memory_list` can also return errors such as `INVALID_PARAMS`, `WORKSPACE_NOT_FOUND`, and `INTERNAL_ERROR`. With the current instructions, an agent could treat those responses as success and proceed to modify `AGENTS.md` or `SOUL.md`, even though memory availability was never established.

Please make the preflight fail closed:

- Continue only when the response contains no `error` field.
- Treat an empty `records` list as successful connectivity.
- On any error, stop without editing the policy file and report the error.
- Add a contract test proving that a non-authorization error also prevents the mutation instructions from being reached.

The same correction is needed in both the Codex and Hermes prompts.

cycterna2222288888-ai:
@toomij99 pushed `8da7660`.

This branch is stacked on #450's base commit and carries its own copy of the same 4 prompt docs, so #450's fail-closed rewrite is mirrored here too (see the #450 thread for the full rationale) — otherwise this branch's copies would regress the moment #450 merges.

Also added the contract test you asked for — a non-authorization error must not let the agent reach the mutation instructions either:

- `tests/unit/mcp/tools/test_memory_list_preflight_contract.py::test_grantless_principal_preflight_fails_with_a_non_auth_error` — binds an authenticated, non-admin principal with no grant on the requested workspace and calls the real `metronix_memory_list`. `resolve_workspace_id` denies it before `require_agent_access` ever runs, so the response comes back `WORKSPACE_NOT_FOUND`, not `AUTH_REQUIRED` — and the memory store is never touched (`build_memory_service_for_workspace` not awaited), same as the existing AUTH_REQUIRED test proves for that path.
- `tests/unit/test_prompt_onboarding_contract.py::test_prompt_2_treats_every_error_as_denial_not_just_auth_required` — doc-text regression guard that the prompt-2 docs name `WORKSPACE_NOT_FOUND`, `INVALID_PARAMS`, `INTERNAL_ERROR`, a non-2xx response, and the tool being unavailable as failures, not only `AUTH_REQUIRED`.

All 51 tests across the 3 affected files pass; `ruff check`/`ruff format --check` clean on the changed files.

toomij99:
=========================== short test summary info ============================
FAILED tests/unit/mcp/tools/test_memory_list_preflight_contract.py::test_admin_principal_preflight_succeeds_via_admin_override - AssertionError: assert 'error' not in {'error': {'code': 'INVALID_PARAMS', 'message': 'A valid X-Agent-Id header is required for authenticated agent tools'}}
FAILED tests/unit/mcp/tools/test_memory_list_preflight_contract.py::test_grantless_principal_preflight_fails_with_a_non_auth_error - AssertionError: assert 'INVALID_PARAMS' == 'WORKSPACE_NOT_FOUND'
  
  - WORKSPACE_NOT_FOUND
  + INVALID_PARAMS
2 failed, 3251 passed, 15 skipped, 3 deselected, 685 warnings in 301.14s (0:05:01)
Error: Process completed with exit code 1.

cycterna2222288888-ai:
@toomij99 Fixed in `94255ac` — the failure is a `main` merge, not a regression in this branch's logic.

**Cause:** `main`'s "enforce MCP agent identity consistency" (`8d66efd`) added a check to the MCP tool wrapper: for an **authenticated** request (`principal is not None`) to any agent-memory tool, the transport `X-Agent-Id` header must be present, valid, and equal to the `agent_id` argument — otherwise `INVALID_PARAMS: "A valid X-Agent-Id header is required for authenticated agent tools"` is returned *before the handler runs*. Unauthenticated local/stdio calls keep the `agent_id`-argument fallback.

Both authenticated tests here bind a principal but never set the transport identity, so they now hit that new guard instead of the authorization paths they exist to check (`admin_override` → clean empty list; grantless → `WORKSPACE_NOT_FOUND`). The shared-key test still passes — no principal, so the guard is skipped.

**Fix:** a real prompt-2 preflight call always carries `X-Agent-Id` (prompt 1 configures it), so the tests now simulate that. Added a `_transport_agent()` context manager — the same shape `tests/integration/mcp/test_agent_access_conformance.py` got in that same `main` commit — that binds `current_agent_id` to the same value passed as `agent_id`. Assertions are unchanged; the `WORKSPACE_NOT_FOUND` / `admin_override` outcomes are restored because the wrapper's identity guard now passes.

`ruff check`/`format --check` clean; `test_memory_list_preflight_contract.py` + `test_prompt_onboarding_contract.py` + `test_agent_access_guard.py` + `test_mcp_tool_wrapper.py` + `test_mcp_agent_memory_auth.py` + `test_mcp_docs.py` + `test_agent_access_conformance.py` — 78 passed locally. CI re-running.

Issue #468, opened 2026-09-07
Confluence incremental sync broken on Cloud: _fetch_incremental calls get_page_by_id, which atlassian.confluence.cloud.Cloud does not have


cycterna2222288888-ai:
### What happened?

`ConfluenceConnector._fetch_incremental` (`src/metronix/connectors/confluence.py`) calls `self._client.get_page_by_id(page_id, expand=...)` for every CQL hit. On a **Confluence Cloud** connection (`*.atlassian.net`) that method does not exist, so every incremental sync fails with:

```
AttributeError: 'Cloud' object has no attribute 'get_page_by_id'
```

`from atlassian import Confluence` in `configure()` builds a dispatch wrapper: for an `*.atlassian.net` / api-gateway URL it sets `self._impl = atlassian.confluence.cloud.Cloud`; otherwise `atlassian.confluence.server.Server`. Verified against the installed `atlassian-python-api` 5.0.4:

```python
>>> from atlassian.confluence.server import Server
>>> from atlassian.confluence.cloud import Cloud
>>> hasattr(Server, "get_page_by_id"), hasattr(Cloud, "get_page_by_id")
(True, False)
>>> hasattr(Server, "get_content"), hasattr(Cloud, "get_content")
(True, True)
```

`get_page_by_id` lives only on `Server` and on the separate v2 `ConfluenceCloud` class (`atlassian/confluence/cloud/cloud.py`), which the connector does not use. The Cloud client (`atlassian.confluence.cloud.Cloud`) exposes `get_content(content_id, **kwargs)` instead — same `GET rest/api/content/{id}` endpoint, same response shape. `get_content` is present on **both** Server and Cloud, so it is the portable call.

The full-sync generator crash (#460) was reproduced against a Server-style mock host, so it never exercised this path. Incremental sync therefore only works against self-hosted Confluence (Server) today; against Confluence Cloud it has never worked since the 5.x client split.

This is an **adjacent finding to #460 and is intentionally out of scope of the #460 fix** (`fix/460-confluence-generator` touches only `_fetch_full` and the response-shape normaliser). Filed so the Cloud incremental path can be fixed with its own regression test.

### Steps to reproduce

1. Create a Confluence source with an `*.atlassian.net` URL and valid credentials:
   `metronix_source_create connector_type=confluence config={url: "https://<org>.atlassian.net/wiki", username: "<email>", api_token: "<token>", space_key: "ENG"}`
[...]

Issue #463, opened 2026-09-07
POST /connections/{id}/test/ returns success:true for wrong credentials — only calls configure(), health_check() is never invoked


cycterna2222288888-ai:
### What happened?

`POST /api/v1/connections/{id}/test/` returns `{"success": true}` for a connection whose stored credentials are certainly wrong (bad token, unreachable host, wrong workspace/space) — as long as the config is syntactically valid.

`test_connection` (`src/metronix/api/routes/connections.py:562`) does exactly one thing with the connector:

```python
connector = registry.create(connector_type)
connection_obj = Connection(id=conn["id"], workspace_id=conn["workspace_id"], connector_type=connector_type)
await connector.configure(connection_obj, conn["config"])          # connections.py:599
# ... then on no exception:
await store.update_connection_status(connection_id, status="active", error_message=None)
return TestConnectionResponse(success=True)
```

`connector.configure()` for every connector is a client constructor — it builds the SDK client object from the decrypted config and does **no network call**. So "test" only proves the config dict has the required keys, not that the credentials work or the host is reachable. The first real sync then fails against the same creds.

`ConnectorInterface.health_check()` (`src/metronix/core/interfaces.py:78` — "Test connectivity to the source. Returns True if reachable.") is implemented on all seven connectors (`confluence.py:233`, `jira.py:177`, `github.py:435`, `notion.py:171`, `gdrive.py:405`, `files.py:53`, `slack_history.py:56`) but is **never called anywhere outside tests** — dead code.

`docs/CONNECTORS.md:33` documents a lifecycle step that does not exist in the code:

```
### Connector Lifecycle
...
2. **Health Check**: Validates connectivity before first sync.
```

**Expected:** `test_connection` should call `connector.health_check()` (a real connectivity probe) and return `success: false` with the failure detail when it fails; or, if `configure()`-only is the intended check, the endpoint response and `CONNECTORS.md` should say so — "config accepted, connectivity not verified" — rather than `success: true` and a documented "Health Check" step.

### Steps to reproduce

1. `docker compose up -d`. Workspace `MTRNIX`.
2. Create a Confluence (or Jira / GitHub / Notion / GDrive) source with a **valid-looking but wrong** config — any URL that accepts a TCP connection, a bogus `api_token`, a `space_key` that does not exist:
   `metronix_source_create connector_type=confluence config={url: "http://localhost:9", username: "x", api_token: "not-a-real-token", space_key: "NOPE"}`
[...]

Issue #459, opened 2026-09-06
Confluence and Jira connector syncs run blocking HTTP on the event loop — a hung source freezes the whole API


cycterna2222288888-ai:
### What happened?

`ConfluenceConnector.fetch()` and `JiraConnector.fetch()` are declared `async def` but perform blocking `requests`-based HTTP calls directly on the event loop — no `asyncio.to_thread`, no async client. Both files still carry a `# TODO: async migration` marker (`connectors/confluence.py:7`, `connectors/jira.py:7`).

While such a sync is waiting on the network, the **entire API event loop is blocked**: MCP tools, `/api/v1` search, memory operations, the OpenAI-compat endpoints, and even `GET /health` stop responding until the connector's HTTP call returns or times out.

Observed: with a Confluence sync stuck on an unresponsive host, `GET /health` and `GET /api/v1/connections/` both time out (`curl -m 5` → exit 28, zero bytes received). The freeze is bounded by `atlassian-python-api`'s default request timeout (~75 s) but recurs for **every** HTTP request the sync makes (space listing + per-page fetch, or each JQL page).

**Consequence chain:**

- The `metronix-core` container healthcheck (`docker-compose.yml`: 5 retries × 10 s on `GET /health`) fails while the loop is blocked → Docker restarts the API container.
- On restart, `recover_interrupted_syncs()` flips the connection `syncing → error`, but autosync re-claims the same source within one poll interval (`next_run_at` NULL = "due now") and the sync hangs again → **restart / crash loop**.
- A single misconfigured or slow Confluence/Jira instance therefore takes down memory + search **for all workspaces**, not just its own sync (the autosync queue and concurrency limit are process-global).

`github`, `notion`, `gdrive` connectors are unaffected — they wrap blocking calls in `asyncio.to_thread` or use an async client.

**Expected:** connector fetches must not block the event loop. Run blocking connector I/O in a worker thread (as GitHub/GDrive already do) or migrate Confluence/Jira to async HTTP.

### Steps to reproduce

1. Create a Confluence source pointing at a host that accepts the TCP connection but never responds (or is very slow):
   `metronix_source_create connector_type=confluence config={url: http://<blackhole>, username: x, api_token: x, space_key: ENG}`
2. Trigger a sync: `metronix_source_sync connection_id=<id>` (or just wait ~60 s for autosync to claim it).
3. From another shell, hit the API while the sync is in flight:
[...]

PR #469, opened 2026-09-07
fix(connectors): iterate the Confluence paginator instead of len()-ing it (#460)

### LINKED ISSUE #460: Confluence full sync always fails: TypeError: object of type 'generator' has no len() (incompatible with atlassian-python-api 5.0.4)
### What happened?

Every full / initial Confluence sync crashes with `TypeError: object of type 'generator' has no len()` at `connectors/confluence.py:109`.

`_fetch_full()` does:

```python
pages = self._client.get_all_pages_from_space(space_key, start=start, limit=limit, expand=expand)
if not pages:
    break
for page in pages:
    ...
if len(pages) < limit:   # line 109 — pages is a generator
    break
```

With the pinned `atlassian-python-api` (`>=5.0.0,<6`, resolved to **5.0.4**), `from atlassian import Confluence` is the v2 `ConfluenceBase`, and `get_all_pages_from_space()` returns `self._get_paged(...)` — a **generator** (its docstring: *"return: A generator object for the data elements"*). `len()` on it raises.

The `for page in pages` loop runs and builds the documents correctly; then `len(pages)` on the now-exhausted generator throws, so `fetch()` returns nothing and the sync ends `status=failed`, `documents_fetched=0`.

**Impact:**

- No new Confluence source can complete its first sync — a fresh connection has `last_synced_at = NULL`, which routes to `_fetch_full()`.
- `force_full=true` also always fails (same path).
- Only `_fetch_incremental()` avoids `len()` on the paginator, and it is unreachable without a prior successful full sync.
- The surfaced error (`object of type 'generator' has no len()`) gives no hint that this is a library-version incompatibility.

**Expected:** materialise the generator into a list (or count yielded items) instead of calling `len()` on the paginator; verify the connector against the API surface actually shipped by the pinned `atlassian-python-api`.

### Steps to reproduce

1. `metronix_source_create connector_type=confluence name=x config={url, username, api_token, space_key: ENG}` — any valid-looking config works; a mock returning `{"results": [...], "size": N}` on `GET /rest/api/content` is enough.
2. `metronix_source_sync connection_id=<id>` — a fresh source has `last_synced_at = NULL` → `_fetch_full`.
3. Poll `metronix_source_list` / `sync_logs`: `status=failed`, `errors=["object of type 'generator' has no len()"]`, `documents_fetched=0`.
4. Repeat with `force_full=true` → same failure.

Direct repro:

```python
from metronix.connectors.confluence import ConfluenceConnector
from metronix.core.models import Connection

c = ConfluenceConnector()
await c.configure(
    Connection(id="t", workspace_id="w", connector_type="confluence"),
    {"url": "http://mock:8899", "username": "u", "api_token": "t", "space_key": "ENG"},
)
await c.fetch("w")   # TypeError: object of type 'generator' has no len()  @ confluence.py:109
```

### Relevant logs / output

```shell
[info]  confluence.fetch.started  since=None workspace_id=w
[debug] confluence.page.processed  title='Onboarding Guide'
[debug] confluence.page.processed  title='Deployment Runbook'
[debug] confluence.page.processed  title='Architecture Overview'
Traceback (most recent call last):
  File ".../metronix/connectors/confluence.py", line 63, in fetch
    return self._fetch_full(workspace_id, base_url, space_key)
  File ".../metronix/connectors/confluence.py", line 109, in _fetch_full
    if len(pages) < limit:
TypeError: object of type 'generator' has no len()

# sync_logs row
status=failed  documents_fetched=0  errors=["object of type 'generator' has no len()"]
```

```shell
$ pip show atlassian-python-api | head -2
Name: atlassian-python-api
Version: 5.0.4
# atlassian/confluence/base.py: def _get_paged(...)  ->  "A generator object for the data elements"
```

### Version / commit

`pr-451` @ `94255ace85c173f631730de3e1c534a13fab4f01`. Same code on `origin/main`.

### Environment

macOS (Darwin 25.6.0), Docker Compose stack (`docker compose up -d`, bundled `docker-compose.yml`), `metronix-core` Python 3.12 container, `atlassian-python-api` 5.0.4.

cycterna2222288888-ai:
Fixes #460.

## What happened

Every full / initial Confluence sync crashed with
`TypeError: object of type 'generator' has no len()` at
`connectors/confluence.py`. `_fetch_full` was written for a library version
where `get_all_pages_from_space(...)` returned a **list** (one page slab), and
paged it manually with `start` / `limit`, checking `len(pages) < limit` to stop.

Verified against the **installed** `atlassian-python-api` 5.0.4 (introspection +
package source, not memory):

- `from atlassian import Confluence` builds a dispatch wrapper — `_impl` is
  `atlassian.confluence.cloud.Cloud` for `*.atlassian.net` URLs, else
  `atlassian.confluence.server.Server`.
- On **both** clients, `get_all_pages_from_space` → `ConfluenceBase._get_paged`,
  which `inspect.isgeneratorfunction` confirms is a **generator** — it walks
  `_links.next` across all pages internally. `len()` on it raises; `if not gen`
  is always false; `for page in gen` drains every page, not `limit`.

### Library version history (why the dependency floor moved)

| Version | `get_all_pages_from_space(...)` returns |
|---|---|
| `<= 3.41` | `list` — one page slab; caller pages with `start`/`limit` |
| `4.0.4 – 4.0.7` | `list` — every page, auto-paginated internally |
| `>= 5.0.0` | **generator** — upstream PR #1616 (*"now return generators instead of dicts"*) + PR #1581 (Cloud/Server split) |

`pyproject.toml` had `>=5.0.0,<6`; bumped the floor to `>=5.0.4` (the tested
build; also includes the 5.0.3 Confluence REST-URL compat and 5.0.4
`get_page_id`-returns-None fixes). `<6` guards the next major, where pagination
already changed once (4 → 5). `uv.lock` pins the exact build.

## Changes

- **`_iter_pages(result)`** helper — normalises `get_all_pages_from_space`'s
  return value (list from `<= 4.x`, generator from `5.x`, or a defensive raw
  `{"results": [...]}` dict) into a single-pass iterator of page dicts.
- **`_fetch_full`** — call `get_all_pages_from_space(space_arg, limit=25,
  expand=...)` once and iterate `_iter_pages(...)`. Removed the manual
  `while True` / `start += limit` loop, the `if not pages` check, and the
  `len(pages)` comparison. Per-page `try/except` kept.
- The pre-5.x inline per-batch **429 retry** (`sleep(4); continue`) is dropped —
  the 5.x paginator raises from inside iteration and cannot be resumed after a
  sleep. A 429 (or any transport error) now fails the full sync; `last_synced_at`
  is not advanced and the next scheduled run retries wholesale. Transparent
  backoff is available at the client level (`backoff_and_retry=True`) as a
  separate follow-up.
- **`pyproject.toml` / `uv.lock`** — pin `atlassian-python-api>=5.0.4,<6`.
- **Tests** (`test_confluence_connector.py`) — `_fetch_full` had **zero**
  coverage, which is why #460 shipped. Added:
  - `TestIterPages` — list, generator, `{"results": [...]}`, dict without
    `results`, `None`.
  - `TestConfluenceFetchFull` — generator result (the #460 case) fully consumed
    with exactly one client call; list result still works; empty generator → 0
    docs; dict result unwrapped; one malformed page in the stream does not abort
    the batch; empty `space_key` passes `None`; a 429 mid-stream propagates.

## Out of scope / adjacent findings (not in this PR)

- **#468** — `_fetch_incremental` calls `get_page_by_id`, which
  `atlassian.confluence.cloud.Cloud` does not have (only `Server` and the v2
  client do; Cloud has `get_content`). Confluence **incremental** sync has never
  worked on Cloud since the 5.x split. Separate fix + test. `_fetch_incremental`
  is untouched here — its `cql()` usage returns a dict on both clients in 5.x and
  is unaffected by the generator change.
- **health_check** — `get_all_spaces` is also a lazy generator now, so
  `health_check` never issues its probe request. `health_check` is dead code in
  production today; tracked in **#463**.
- **#459** — Confluence/Jira blocking HTTP on the event loop. Unrelated.
- **Jira** — not affected by #460. `enhanced_jql` returns a dict
  (`GET /rest/api/3/search/jql`) in 5.x; the connector's token pagination is
  correct.

## Test plan

- [x] `mypy src/metronix/connectors/confluence.py` — 9 errors, all pre-existing
  (`self._client` typed `None`, `Document(**{...})` splat); baseline was 10,
  this PR removes one. CI does not gate mypy.
- [x] `tests/unit/test_confluence_connector.py` — 30 passed (12 new)
- [x] connector + ingestion-sync unit suites — green
- [x] the 3 `test_connections_sync.py::test_run_connection_sync_*` errors are
  pre-existing (need a live Postgres; in the repo's known-failing set)
- [ ] CI `pytest (unit, with services)` — green (runs those DB tiers)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,7 +53,11 @@ dependencies = [
 
 [project.optional-dependencies]
 connectors = [
-    "atlassian-python-api>=5.0.0,<6",
+    # 5.x split Confluence into Cloud/Server clients and made get_all_* return
+    # generators (upstream PR #1616); the connector depends on that contract plus
+    # the 5.0.3/5.0.4 Confluence REST fixes. <6 guards the next major, where
+    # pagination behaviour changed once before (4 -> 5). uv.lock pins the build.
+    "atlassian-python-api>=5.0.4,<6",
     "notion-client>=2.2,<3",
     "PyGithub>=2.5,<3",
     "google-api-python-client>=2.198.0,<3",
--- a/src/metronix/connectors/confluence.py
+++ b/src/metronix/connectors/confluence.py
@@ -9,7 +9,9 @@ from __future__ import annotations
 
 import contextlib
 import time
+from collections.abc import Iterator
 from datetime import datetime
+from typing import Any
 
 import structlog
 
@@ -19,6 +21,28 @@ from metronix.core.interfaces import ConnectorInterface
 from metronix.core.models import Connection, Document
 
 logger = structlog.get_logger()
+
+
+def _iter_pages(result: Any) -> Iterator[dict[str, Any]]:
+    """Normalise ``get_all_pages_from_space``'s return value to an iterator of pages.
+
+    ``atlassian-python-api`` changed this method's contract across majors:
+
+    * ``<= 3.41`` — a ``list``: one page slab of up to ``limit`` items; the
+      caller paginated with ``start`` / ``limit``.
+    * ``4.0.4 - 4.0.7`` — a ``list``: every page, auto-paginated internally.
+    * ``>= 5.0.0`` — a **generator** that walks ``_links.next`` across all pages
+      (upstream PR #1616). ``len()`` on it raises
+      ``TypeError: object of type 'generator' has no len()`` — issue #460.
+
+    A raw ``{"results": [...]}`` dict is also tolerated defensively. The result
+    is always iterated exactly once, so a generator is safe to pass through.
+    """
+    if result is None:
+        return iter(())
+    if isinstance(result, dict):
+        return iter(result.get("results", []))
+    return iter(result)
 
 
 class ConfluenceConnector(ConnectorInterface):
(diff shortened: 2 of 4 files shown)


toomij99 (approved):
Reviewed: no actionable findings.

Issue #470, opened 2026-09-07
Benchmark env loader (_parse_env_file) strips no inline comments and reads the repo-root .env — diverges from the Compose env_file parser


cycterna2222288888-ai:
### What happened?

The benchmark env loaders parse env files with a hand-rolled parser that strips **no** inline comments and also reads the repo-root `.env` — so its behaviour diverges from Docker Compose's `env_file` parser, and it inherits the `.env.example` inline-comment problem (#458) in a stricter form.

`benchmarks/locomo/scripts/env_config.py` and `benchmarks/longmemeval/scripts/env_config.py` both define:

```python
def _parse_env_file(path: Path) -> dict[str, str]:
    ...
    for raw_line in path.read_text(encoding="utf-8").splitlines():
        line = raw_line.strip()
        if not line or line.startswith("#"):
            continue
        if "=" not in line:
            continue
        key, _, value = line.partition("=")
        values[key.strip()] = value.strip().strip('"').strip("'")
    return values
```

There is no `#` handling anywhere. Consequences:

- `KEY=value   # note` → value `value   # note` (Compose's `env_file` parser cuts at the first ` #` and would give `value`).
- `KEY=   # note` → value `# note` (same end state as #458, different mechanism).
- Only a whole-line leading `#` is treated as a comment.

And both loaders read the **repo-root `.env`** in addition to their own `.env.benchmark`:

```python
# locomo
REPO_ENV_PATH = REPO_ROOT / ".env"
def load_dotenv() -> None:
    for key, value in _parse_env_file(REPO_ENV_PATH).items():
        os.environ.setdefault(key, value)
    ...
```

So a benchmark run parses the same repo-root `.env` (copied from `.env.example`) with a parser that is *less* forgiving than Compose's. Today the only repo-`.env` key these consume is `METRONIX_MCP_API_KEY`, which `.env.example` ships as a bare line — so there is no live breakage right now — but the parser is a latent trap and is inconsistent with how the rest of the stack reads the same file.

The two `.env.benchmark.example` files themselves are clean (comments on their own lines), so this is about the parser, not the shipped templates.

### Expected

`_parse_env_file` should either:

- strip inline comments the same way Compose does (cut at the first ` #` on unquoted values), so the two parsers agree on any file they might both read; or
- be replaced with a shared helper, or `python-dotenv`, so there is one env-parsing behaviour in the repo.

[...]

PR #471, opened 2026-09-07
fix(env): move .env.example inline comments to their own line (#458)

### LINKED ISSUE #458: .env.example inline comments become variable values via Compose env_file (~7 vars incl. QDRANT_API_KEY, AUTH_PASSWORD, LLM_FALLBACK_PROVIDER)
### What happened?

`.env.example` (shipped in the repo; `make setup` runs `cp -n .env.example .env`; `docker-compose.yml` loads
it via `env_file: .env`) has variable lines of the form `KEY=<spaces># explanatory comment`. Docker Compose's
`env_file` parser does not strip the trailing inline comment, so the comment text becomes the variable's
value in the container, and `Settings` picks it up.

Affected lines in `.env.example`:

```
37:  QDRANT_API_KEY=                       # only for cloud / authenticated Qdrant
53:  AUTH_PASSWORD=                        # required to seed admin@metronix.local; generate with: openssl rand -base64 32
76:  OLLAMA_REQUEST_TIMEOUT=               # seconds; empty = library default
96:  METRONIX_FRESHNESS_LLM_PROVIDER=      # set with API_BASE_URL -> auto "custom"
155: LLM_FALLBACK_PROVIDER=                # fallback provider if primary fails
253: LLM_PROVIDER_API_KEY=                 # optional; blank = no auth on the endpoint
254: LLM_PROVIDER_MODEL=                   # required when LLM_PROVIDER=custom
```

Verified in the running container:

```
$ docker compose exec metronix-core sh -c 'echo "[$LLM_FALLBACK_PROVIDER]"'
[# fallback provider if primary fails]

$ docker compose exec metronix-core sh -c 'echo "[$QDRANT_API_KEY]"'
[# only for cloud / authenticated Qdrant]

$ docker compose exec metronix-core python -c "from metronix.core.config import get_settings as g; print(repr(g().llm_fallback_provider))"
'# fallback provider if primary fails'

$ ... g().qdrant_api_key       -> '# only for cloud / authenticated Qdrant'
$ ... g().llm_provider_model   -> '# required when LLM_PROVIDER=custom'
$ ... g().freshness_llm_provider -> '# set with API_BASE_URL -> auto "custom"'
```

Each of these is intended to be empty by default.

### Observed impact

- `LLM_FALLBACK_PROVIDER`: when the primary LLM (ollama) timed out, `src/metronix/llm/provider.py:74`
  `fallback_name = settings.llm_fallback_provider` is the truthy string `'# fallback provider if primary
  fails'`, so `create_provider(...)` is attempted with that name:

  ```
  warning fallback_provider_creation_failed
    provider='# fallback provider if primary fails'
    error='Unknown LLM provider: # fallback provider if primary fails. Available: deepseek, openrouter, ollama, custom'
  ```

  Instead of "no fallback configured → degrade", the fallback path always fails and logs an error.

- `QDRANT_API_KEY`: a non-empty junk value makes the Qdrant client log
  `UserWarning: Api key is used with an insecure connection` on every startup
  (`src/metronix/storage/memory_qdrant.py:65`), and would send that string as the API key to an authenticated
  Qdrant endpoint.

- `AUTH_PASSWORD` resolves to the comment text; it is documented as the seed password for
  `admin@metronix.local`. Not separately exercised here (`AUTH_ENABLED=false` in this environment).

- `OLLAMA_REQUEST_TIMEOUT` resolves to `# seconds; empty = library default` (a non-numeric string).

### Expected

Empty values in `.env.example` should resolve to empty. Either move the explanatory text to its own comment
line above each variable, or strip inline comments when loading.

### Steps to reproduce

1. `cp .env.example .env` (or `make setup`).
2. `docker compose up -d`.
3. `docker compose exec metronix-core sh -c 'echo "[$LLM_FALLBACK_PROVIDER]" "[$QDRANT_API_KEY]"'`
   → `[# fallback provider if primary fails] [# only for cloud / authenticated Qdrant]`
4. `docker compose exec metronix-core python -c "from metronix.core.config import get_settings as g; \
   print(repr(g().llm_fallback_provider), repr(g().qdrant_api_key), repr(g().llm_provider_model))"`
5. Cause the primary LLM to fail (e.g. stop the `ollama` container) and issue a request that needs the LLM
   → log shows `fallback_provider_creation_failed provider='# fallback provider if primary fails'`.

### Relevant logs / output

```
raw env  LLM_FALLBACK_PROVIDER = [# fallback provider if primary fails]
raw env  QDRANT_API_KEY        = [# only for cloud / authenticated Qdrant]

get_settings().llm_fallback_provider  = '# fallback provider if primary fails'
get_settings().qdrant_api_key         = '# only for cloud / authenticated Qdrant'
get_settings().llm_provider_model     = '# required when LLM_PROVIDER=custom'
get_settings().freshness_llm_provider = '# set with API_BASE_URL -> auto "custom"'

warning fallback_provider_creation_failed
  provider='# fallback provider if primary fails'
  error='Unknown LLM provider: # fallback provider if primary fails. Available: deepseek, openrouter, ollama, custom'
```

### Version / commit

`main` @ `ca50e5a`, local `docker compose` stack.

### Environment

macOS 15.6, Docker 29.6.2 / Compose v2. `.env` from `.env.example` via `make setup`. `docker-compose.yml`
`metronix-core` service loads it with `env_file: .env`.

cycterna2222288888-ai:
Fixes #458.

## What happened

`docker-compose.yml` loads `.env` via `env_file:`, which Compose parses with
`compose-spec/compose-go`'s `dotenv` package. For an unquoted line it:

1. strips **leading** whitespace from the value (`locateKeyName`:
   `strings.TrimLeftFunc(src[offset:], isSpace)`), then
2. removes an inline comment only at the first literal `" #"` — a space
   immediately followed by `#` (`extractVarValue`: `strings.Cut(value, " #")`),
   then trims trailing whitespace.

So `KEY=value   # note` → `value`, but for an intended-**empty** value
`KEY=   # note` the leading spaces are removed first, the value now *starts*
with `#`, there is no `" #"` to cut, and the whole `# note` becomes the value.

`.env.example` had **7 lines** that hit this (value resolves to `# ...`):
`QDRANT_API_KEY`, `AUTH_PASSWORD`, `OLLAMA_REQUEST_TIMEOUT`,
`METRONIX_FRESHNESS_LLM_PROVIDER`, `LLM_FALLBACK_PROVIDER`,
`LLM_PROVIDER_API_KEY`, `LLM_PROVIDER_MODEL` — plus **50 more** `KEY=value  # …`
lines that currently work only because a space precedes the `#`.

Observed downstream (from #458): `LLM_FALLBACK_PROVIDER` →
`Unknown LLM provider: # fallback provider…` on every primary-LLM failure;
`QDRANT_API_KEY` → "insecure connection" warning + a junk key sent to an
authenticated Qdrant; `AUTH_PASSWORD` → the seeded admin password is comment
text.

Blast radius: `docker compose up -d` (`cp .env.example .env`) breaks all 7.
`install.sh` rewrites lines for keys it sets, but of the 7 only
`LLM_PROVIDER_{API_KEY,MODEL}` are set (and only on `--mode answers`), so the
other 5 break via `install.sh` too. `install.sh` already carried a workaround
(`example_val()` stripped inline comments so its placeholder check wouldn't
compare against comment text).

## Changes

- **`.env.example`** — every inline comment (all 57 lines, not just the 7 that
  break) moved to its own line directly above the assignment. This is the only
  form that is correct under every parser (compose-go, python-dotenv,
  `set -a; source`, `docker run --env-file`, and the benchmark loader), and it
  matches the style the ZONE 3 section and the `.env.benchmark.example` files
  already use. Header gains a short note explaining the convention.
- **`install.sh`** — drop the now-unnecessary `sed 's/[[:space:]]#.*//'` from
  `example_val()`; the value is taken verbatim (trailing whitespace still
  trimmed as a guard).
- **`tests/unit/test_env_example.py`** (new) — parses `.env.example`
  line-by-line, emulates the compose-go `dotenv` rules, and asserts (a) every
  non-comment line is `KEY=VALUE`, (b) no line carries an inline `#` at all,
  (c) no value resolves to comment text under Compose, (d) the 7 keys from
  #458 resolve to empty.

## Adjacent finding (not in this PR)

**#470** — the benchmark env loader
(`benchmarks/*/scripts/env_config.py::_parse_env_file`) strips no inline
comments *at all* and also reads the repo-root `.env`, so it diverges from the
Compose `env_file` parser. The `.env.benchmark.example` templates themselves are
clean; the parser is the latent trap. Filed separately.

## Test plan

- [x] `tests/unit/test_env_example.py` — 3 passed
- [x] `tests/unit/test_mcp_docs.py` — 40 passed (it reads `.env.example`, incl.
  the parametrised `PUBLIC_MCP_MODE_DOCS` check)
- [x] `bash -n install.sh` — OK; `tests/installer/test_install.sh` — 49 passed /
  3 failed, **identical on the clean tree** (the 3 are a docker-volume-removal
  path the sandbox can't stub — pre-existing)
- [x] a script emulation of the compose-go parser over the new `.env.example`
  reports 128 assignment lines, 0 inline-comment leaks, 0 `#`-after-`=`
- [ ] CI: `installer` (shellcheck + `bash -n`) and `pytest (unit, with services)`

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -10,6 +10,11 @@
 #
 # All vars mirror src/metronix/core/config.py (Settings) plus a few read directly
 # via os.environ (MCP server, query-expansion LLM, admin cleanup, entity resolver).
+#
+# Comments go on their own line, above the value they annotate. An inline
+# "# ..." after a value is NOT portable: Docker Compose's env_file parser only
+# strips it when a space precedes the '#' AND the value is non-empty, so
+# `KEY=   # note` resolves to the literal value "# note" (issue #458).
 # =============================================================================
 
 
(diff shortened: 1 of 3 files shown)


toomij99 (approved):
Reviewed the environment-template rewrite, installer parsing change, and regression tests. No actionable findings; all required checks pass.

Issue #457, opened 2026-09-05
No user-facing document delete; DELETE /connections orphans documents; internal-primitive delete leaves the PG row active and orphan entities


cycterna2222288888-ai:
### What happened?

**Part 1 — there is no user-facing way to delete a knowledge-base document, and `DELETE /connections/{id}`
leaves the documents behind.**

- No `DELETE` route under `/api/v1/knowledge` — `src/metronix/api/routes/knowledge.py` exposes only
  `GET /records` and `POST /store`.
- No `metronix_*` MCP tool deletes a KB document (22 tools; none remove a document).
- `src/metronix/api/routes/documents.py` exposes only `GET /{document_id}/history`.
- No connector reconciliation for upstream-deleted documents — no `deleted_source_ids` / `removed_from_source`
  / prune path in `src/metronix/ingestion/`. A Jira/Confluence/etc. document deleted at the source stays in
  Metronix indefinitely. The only pipeline delete (`src/metronix/ingestion/pipeline.py`, ~line 210) is the
  incremental *re-ingest* path: same `source_id`, changed content → drop old chunks, add new.
- `DELETE /api/v1/connections/{connection_id}/` (`src/metronix/api/routes/connections.py:532`) runs:

  ```python
  await channel_manager.stop_channel(connection_id)
  await store.delete_connection(connection_id)
  ```

  It deletes the connection row + encrypted credentials only. No `delete_by_doc_labels`, no `raw_documents` /
  Qdrant / Neo4j cleanup. Every document that connection ingested keeps its `raw_documents` row (now with a
  dangling `connection_id`), its Qdrant points, and its Neo4j nodes.
- The only cleanup that removes documents is workspace-scoped: `DELETE /api/v1/admin/cleanup/workspace/{id}`
  and `DELETE /api/v1/admin/cleanup/all`.

**Part 2 — deleting a document with the internal primitives (what incremental sync uses) leaves residue in
Postgres and Neo4j.**

Running the two primitives incremental sync uses to drop a document —
`get_hybrid_store().delete_by_doc_labels([label])` and
`metronix.storage.graph_ops.delete_document_node(label, workspace_id)` — on an existing document
(`QA-MULTILANG`, which had extracted entities `超时`, `Error code E_TIMEOUT 504`, `Helios`, `Aurora`):

| Store | State after both primitives |
|---|---|
| Qdrant points for `doc_label` | 0 — removed |
[...]

PR #473, opened 2026-09-07
fix(sync): delete-before-add on connector re-ingest, skip when unchanged (#461)

### LINKED ISSUE #461: Connector sync re-ingests unchanged documents into Qdrant without delete-before-add — duplicate chunks, inflated documents_new, PG/index divergence
### What happened?

When a connector sync fetches documents that are byte-identical to what is already stored, `run_connection_sync` still re-ingests them into Qdrant **without removing the previous chunks**, so every affected document gets a second (third, …) full copy of its chunks in the vector store.

Two compounding issues in `connectors/connection_sync.py`:

1. **Phase 2 fallback** (`connection_sync.py:285`):

   ```python
   if upsert_result and upsert_result.get("changed_source_ids"):
       docs_to_ingest = [d for d in documents if d.source_id in changed_ids]
   else:
       docs_to_ingest = documents          # also hit when NOTHING changed
   ```

   `changed_source_ids` is empty both when the raw-document upsert failed **and** when every fetched document is unchanged. The `else` branch re-ingests the entire fetched batch in the second case.

2. **No delete-before-add.** The ingest call (`connection_sync.py:289`) passes no `incremental` flag, and `ingest_documents(incremental: bool = False)` only deletes prior chunks by `doc_label` when `incremental=True` (`ingestion/pipeline.py:208-210`). The connector sync path therefore never deletes old chunks before writing new ones.

`raw_documents` (PostgreSQL, source of truth) dedupes correctly on `(workspace_id, connector_type, source_id)` and stays at the right count — only the Qdrant index grows. **PG and the vector index diverge.**

**Observed impact:**

- Chunk count per `doc_label` grows by one full copy per qualifying sync — measured **3 → 6 → 9** for 3 documents.
- `metronix_search_fast` returns the same document as several separate hits with near-identical scores (one physical passage occupying N result slots), degrading search / RAG context.
- `sync_logs.documents_new` reports the re-ingested count (e.g. `3`) when 0 documents were new or changed; `qdrant_chunks` counts the duplicate writes.
- Autosync makes this **recurring** for any connector whose incremental `since` filter re-returns unchanged rows.

**Triggers observed / expected:**

- `force_full=true` on a source with no upstream changes.
- A second connector pointed at the same source — its first sync re-fetches everything the first connector already ingested.
- Cursor reset, or minute-precision `since` re-emitting boundary rows (CQL / JQL date filters are minute-resolution).

**Expected:** skip ingest entirely when nothing changed; and when re-ingest is intended, delete the prior chunks for those `doc_label`s first (pass `incremental=True`, or make the connector path delete-before-add).

### Steps to reproduce

1. Create a Confluence source against a mock serving 3 pages. Seed `connections.last_synced_at` to a past date so the working incremental path is used: `UPDATE connections SET last_synced_at='2026-08-01T00:00:00+00' WHERE id=<A>`. Sync → 3 docs, 3 Qdrant chunks (`type=confluence`).
2. Create a **second** Confluence source against the same mock / space. Seed its `last_synced_at` to the same past date. Sync it.
3. Count Qdrant chunks: `POST /collections/mem_docs_hybrid/points/count {exact:true, filter:{must:[{key:"type",match:{value:"confluence"}}]}}` → **6** (was 3). Each `doc_label` now has 2 identical chunks. `raw_documents` still has 3 rows.
4. `sync_logs` for the second sync: `documents_new=3, qdrant_chunks=3` — despite 0 new / 0 changed.
5. Reset a source's cursor and sync again → count goes **6 → 9**.
6. `metronix_search_fast query="deployment runbook"` → the same "Deployment Runbook" document returned 3× as separate results.

(Confluence's own `force_full` can't be used for the simplest form of step 2 because of the generator crash in `_fetch_full` — filed separately — hence the cursor-seed setup. The bug itself is connector-agnostic: it is in `run_connection_sync` + `ingest_documents`.)

### Relevant logs / output

```shell
[info] sync.raw_documents.persisted  new=0 unchanged=3 updated=0
[info] sync.logged  status=success documents_new=3 qdrant_chunks=3     # 0 actually new/changed

$ curl -s .../collections/mem_docs_hybrid/points/count \
    -d '{"exact":true,"filter":{"must":[{"key":"type","match":{"value":"confluence"}}]}}'
{"result":{"count":3}}   # after connector A
{"result":{"count":6}}   # after connector B (same 3 docs, unchanged)
{"result":{"count":9}}   # after a cursor-reset re-sync
# per doc_label after step 5:  1001 -> 3,  1002 -> 3,  1003 -> 3

# metronix_search_fast "deployment runbook"
#   hit 1: doc_label 1002  score 0.03279
#   hit 2: doc_label 1002  score 0.03226
#   hit 3: doc_label 1002  score 0.03175

# code
connectors/connection_sync.py:285   else: docs_to_ingest = documents
connectors/connection_sync.py:289   result = await ingest_documents(docs_to_ingest, workspace_id,
                                        connector_type, source_role=..., skip_graph=True)  # no incremental=True
ingestion/pipeline.py:133            async def ingest_documents(..., incremental: bool = False, ...)
ingestion/pipeline.py:208            if incremental and doc.source_id:
ingestion/pipeline.py:210                deleted = await store.delete_by_doc_labels([doc.source_id])
```

### Version / commit

`pr-451` @ `94255ace85c173f631730de3e1c534a13fab4f01`. Same code on `origin/main`.

### Environment

macOS (Darwin 25.6.0), Docker Compose stack (`docker compose up -d`, bundled `docker-compose.yml`), `metronix-core` Python 3.12 container, Qdrant `mem_docs_hybrid` collection, `AUTH_ENABLED=false`.

cycterna2222288888-ai:
Fixes #461.

## What happened

`run_connection_sync` Phase 2 (`connectors/connection_sync.py`) called
`ingest_documents(...)` **without `incremental=True`**, and its fallback branch
re-ingested the whole fetched batch whenever `changed_source_ids` was empty:

```python
if upsert_result and upsert_result.get("changed_source_ids"):
    docs_to_ingest = [d for d in documents if d.source_id in changed_ids]
else:
    docs_to_ingest = documents            # <- re-ingest everything
result = await ingest_documents(docs_to_ingest, ...)   # <- no incremental flag
```

Qdrant point ids are random UUIDs (`AsyncQdrantVectorStore.add_document`:
`qdrant_id = str(uuid.uuid4())`), so a document re-ingested without first
dropping its old points gets a **second full copy of every chunk** — measured
3 → 6 → 9. The dedup index can't stop it: `DeduplicationIndex.check_and_add`
only flags a near-duplicate **from a different `doc_label`** ("Same-document
chunks are not flagged (handled by delete-before-reingest)"). The only
delete-before-add path is `ingest_documents(incremental=True)` →
`delete_by_doc_labels([doc.source_id])`, which Phase 2 never triggered.

`raw_documents` dedups on `content_hash` and stays at N rows; only the Qdrant
index grows — PG and the index diverge. `changed_source_ids` is empty (→ the
fallback) whenever every fetched doc is unchanged and already `qdrant_synced`:
**force_full**, a **second connector on the same source** (`raw_documents` is
keyed `(workspace_id, connector_type, source_id)`, not `connection_id`), a
**cursor reset**, and **CQL/JQL minute-boundary re-emits**.

## Changes

**`connectors/connection_sync.py` — Phase 2**
- Always `incremental=True` on the `ingest_documents` call. `delete_by_doc_labels`
  is a cheap no-op for a brand-new doc (scroll finds nothing → 0), so this is
  safe for `new` docs and correct for `updated` ones (whose stale chunks were
  previously left behind too).
- Three explicit branches instead of the `if/else`:
  - `upsert_result is None` (persist raised) → **skip ingest**, add the
    sanitized persist error so the sync is not reported `success` (was: full
    re-ingest onto an unknown PG state, marked `success`).
  - `changed_source_ids` non-empty → ingest only those docs (unchanged).
  - otherwise (nothing changed) → **skip ingest** — Qdrant already holds these
    chunks (was: re-ingest the whole batch → the #461 duplication).

**`connectors/connection_sync.py` — Phase 3b (graph rebuild)** — addressing the
review: `incremental=True` also fires `_delete_graph_node` inside
`ingest_documents` (delete-before-add), but the ingest runs `skip_graph=True`
so nothing rebuilds inline, and the delete never touches `graph_synced`. For
`new` / content-changed docs that's fine — `upsert_raw_documents` already set
`graph_synced=false`, so the Phase 4 sweeper rebuilds them. For a **sidecar-only
refresh** (#440 — content unchanged, so `graph_synced` stays `true`, but the
`source_id` is still returned in `changed_source_ids`) the sweeper would never
pick the doc back up → the graph node stays deleted until the next content
change.

Phase 3b now clears `graph_synced` for the re-ingested `source_id`s via the new
`PostgresStore.mark_documents_graph_unsynced_by_source`, so Phase 4 rebuilds
them. It's a no-op for the rows `upsert_raw_documents` already left unsynced, so
only the sidecar rows actually flip. Local to `connection_sync` — the shared
`ingest_documents` pipeline and its other callers are untouched.

**`mcp/sync.py`** — the MCP-client connector had the same class of bug:
`incremental=not force_full` inverted it exactly on the `force_full` path
(force_full → `docs_to_ingest = all_docs`, `incremental=False` → duplicate).
Now `incremental=True`. (No graph concern here — this path runs
`ingest_documents` with `skip_graph=False`, so the node is rebuilt inline.)

**Behaviour note:** with this PR a connector **`force_full`** re-fetches from
source and refreshes `raw_documents.fetched_at`, but **no longer rewrites Qdrant
for documents whose content is unchanged** — the index already holds those
chunks. This is intentional (#461); a genuine "rebuild the index from scratch"
still needs a reindex path, not `force_full`.

**Tests** (`test_connections_sync.py`, mocked store — no live PG):
1. `run_connection_sync` passes `incremental=True` to `ingest_documents`.
2. an unchanged re-fetch (`changed_source_ids == []`) does **not** call
   `ingest_documents` and finishes `status=success` / connection `active`.
3. an `upsert_raw_documents` failure does **not** call `ingest_documents`,
   finishes non-`success` / connection `error`, and carries the sanitized
   persist error into the `sync_logs` row.
4. **(review regression)** a sidecar-drift shape (`unchanged=1`,
   `changed_source_ids=["J-1"]`) re-ingests with `skip_graph=True` **and** calls
   `mark_documents_graph_unsynced_by_source(source_ids=["J-1"])` so Phase 4
   rebuilds the node the incremental delete dropped.

`test_mcp_client.py::test_sync_server_force_full` now also asserts
`incremental=True`, guarding the `incremental=not force_full` inversion from
regressing.

## Adjacent — not fixed here

`AsyncQdrantVectorStore.delete_by_doc_labels` returns `len(scroll(limit=100))`,
so the deleted count is under-reported past 100 points on a single `doc_label`
(the `client.delete(points_selector=filt)` itself is unbounded). Metric/log
only; noted in #457.

## Test plan

- [x] `mypy connection_sync.py mcp/sync.py storage/postgres.py` — error counts
  unchanged vs `main` (pre-existing `mcp/sync.py` no-any-return; the 22
  `postgres.py` errors are all pre-existing `changed_source_ids`/`dict` noise)
- [x] the 4 `test_phase2_*` tests + `test_sync_server_force_full` — pass locally
  (mocked store, no PG)
- [x] `test_connection_sync_refactor.py`, `test_connection_sync_cursor.py`,
  `test_ingest_sync_shared.py`, `test_mcp_client.py` — pass
- [x] `test_raw_documents.py` — 5 pre-existing failures (SQLite test schema
  lacks the `graph_failed` column, per #440); identical on `main`
- [ ] CI `pytest (unit, with services)` — runs the live-PG
  `test_connections_sync.py` tiers

### DIFF
--- a/src/metronix/connectors/connection_sync.py
+++ b/src/metronix/connectors/connection_sync.py
@@ -282,6 +282,7 @@ async def run_connection_sync(
 
         # Phase 1: Persist raw documents to PostgreSQL (source of truth)
         upsert_result: dict[str, Any] | None = None
+        upsert_error: str | None = None
         try:
             upsert_result = await store.upsert_raw_documents(
                 workspace_id=workspace_id,
@@ -298,6 +299,7 @@ async def run_connection_sync(
                 unchanged=upsert_result["unchanged"],
             )
         except Exception as e:
+            upsert_error = sanitize_error(str(e))
             logger.warning("sync.raw_documents.error", error=str(e))
 
         # Phase 1b: Enqueue KB freshness jobs for changed docs (PROJ-313).
@@ -348,7 +350,8 @@ async def run_connection_sync(
             # Phase 1 persist failed — we do not know what changed. Writing to
             # Qdrant on top of an unknown PG state is worse than skipping.
             docs_to_ingest = []
-            errors_list.append("raw_documents persist failed; Qdrant ingest skipped")
+            msg = "raw_documents persist failed; Qdrant ingest skipped"
+            errors_list.append(f"{msg}: {upsert_error}" if upsert_error else msg)
         elif upsert_result.get("changed_source_ids"):
             changed_ids = set(upsert_result["changed_source_ids"])
             docs_to_ingest = [d for d in documents if d.source_id in changed_ids]
@@ -396,6 +399,28 @@ async def run_connection_sync(
                     )
             except Exception as e:
                 logger.warning("sync.mark_synced.error", error=str(e))
+
+            # Phase 3b: The ingest above ran with ``incremental=True``, which
+            # deletes each re-ingested doc's graph node (delete-before-add in
+            # ``ingest_documents``) but not its ``graph_synced`` flag — and it
+            # ran with ``skip_graph=True``, so nothing rebuilt the node inline.
+            # For new / content-changed docs ``upsert_raw_documents`` already
+            # set ``graph_synced=false``; a sidecar-only refresh (#440) leaves
+            # it ``true``, so Phase 4 below would never rebuild the node that
+            # was just deleted — permanent graph loss until the next content
+            # change (#461). Force the re-ingested source ids unsynced so the
+            # graph sweeper picks them up. Idempotent for the already-false
+            # rows.
+            try:
+                ingested_source_ids = [d.source_id for d in docs_to_ingest if d.source_id]
+                if ingested_source_ids:
+                    await store.mark_documents_graph_unsynced_by_source(
+                        workspace_id=workspace_id,
+                        connector_type=connector_type,
+                        source_ids=ingested_source_ids,
+                    )
+            except Exception as e:
+                logger.warning("sync.mark_graph_unsynced.error", error=str(e))
         elif upsert_result is None:
             status = "failed"
         else:
--- a/src/metronix/connectors/connection_sync.py
+++ b/src/metronix/connectors/connection_sync.py
@@ -339,8 +339,17 @@ async def run_connection_sync(
                     payload={"connector_type": connector_type, "source_id": src_id},
                 )
 
-        # Phase 2: Ingest into Qdrant (only new/updated docs, skip unchanged)
-        if upsert_result and upsert_result.get("changed_source_ids"):
+        # Phase 2: Ingest into Qdrant. Every ingest is delete-before-add per
+        # doc_label (``incremental=True``): Qdrant point ids are random UUIDs,
+        # so re-ingesting a document without first dropping its old points
+        # duplicates every chunk (#461). ``delete_by_doc_labels`` is a cheap
+        # no-op for a genuinely new document that has no points yet.
+        if upsert_result is None:
+            # Phase 1 persist failed — we do not know what changed. Writing to
+            # Qdrant on top of an unknown PG state is worse than skipping.
+            docs_to_ingest = []
+            errors_list.append("raw_documents persist failed; Qdrant ingest skipped")
+        elif upsert_result.get("changed_source_ids"):
             changed_ids = set(upsert_result["changed_source_ids"])
             docs_to_ingest = [d for d in documents if d.source_id in changed_ids]
             logger.info(
@@ -350,7 +359,10 @@ async def run_connection_sync(
                 skipped=len(documents) - len(docs_to_ingest),
             )
         else:
-            docs_to_ingest = documents
+            # Nothing new or changed — Qdrant already holds these chunks;
+            # re-ingesting would duplicate them (#461).
+            docs_to_ingest = []
+            logger.info("sync.nothing_to_ingest", total=len(documents))
 
         if docs_to_ingest:
             result = await ingest_documents(
@@ -359,6 +371,7 @@ async def run_connection_sync(
                 connector_type,
                 source_role=connector.source_role,
                 skip_graph=True,
+                incremental=True,
             )
             documents_new = result.documents_new
             documents_updated = result.documents_updated
@@ -383,6 +396,8 @@ async def run_connection_sync(
                     )
             except Exception as e:
                 logger.warning("sync.mark_synced.error", error=str(e))
+        elif upsert_result is None:
+            status = "failed"
         else:
             status = "success"
 
--- a/src/metronix/mcp/sync.py
+++ b/src/metronix/mcp/sync.py
@@ -142,7 +142,10 @@ class MCPSyncManager:
             docs_to_ingest,
             workspace_id,
             connector_type=f"mcp:{config.name}",
-            incremental=not force_full,
+            # Always delete-before-add per doc_label. On force_full the whole
+            # batch is re-ingested; without the delete, random-UUID Qdrant
+            # points make that a second full copy of every chunk (#461).
+            incremental=True,
         )
 
         logger.info(
--- a/src/metronix/storage/postgres.py
+++ b/src/metronix/storage/postgres.py
@@ -1544,6 +1544,47 @@ class PostgresStore:
                 },
             )
 
+    async def mark_documents_graph_unsynced_by_source(
+        self,
+        workspace_id: str,
+        connector_type: str,
+        source_ids: list[str],
+    ) -> None:
+        """Force ``graph_synced=false`` for the given source docs.
+
+        Used by the connector sync after a ``skip_graph=True`` incremental
+        Qdrant re-ingest: that path deletes each re-ingested doc's graph node
+        (delete-before-add in ``ingest_documents``) without touching
+        ``graph_synced``. For a sidecar-only refresh (#440) the row is left
+        ``graph_synced=true``, so the decoupled graph sweeper would never
+        rebuild the node it just lost (#461). Clearing the flag re-enqueues
+        the doc for extraction; a no-op for rows already ``false``.
+        """
+        if not source_ids:
+            return
+
+        logger.info(
+            "postgres.raw_documents.mark_graph_unsynced_by_source",
+            count=len(source_ids),
+        )
+
+        async with self._engine.begin() as conn:
+            await conn.execute(
+                text("""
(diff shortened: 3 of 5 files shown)


toomij99 (changes requested):
## Request changes

The #461 Qdrant-duplication fix is correct and well-tested: always `incremental=True`, skip ingest when `changed_source_ids` is empty, fail (no ingest) when persist fails, and the MCP `force_full` twin-bug (`incremental=not force_full`) is fixed. The three new connection_sync tests look good and CI is green.

### Blocking: `incremental=True` × `skip_graph=True` × sidecar

`ingest_documents(incremental=True)` deletes the graph node whenever prior Qdrant points exist (`pipeline.py` delete-before-add → `_delete_graph_node`). Content updates reset `graph_synced=false`, so Phase 4 rebuilds. **Sidecar-only drift (#440) does not** — by design graph flags are left alone when content is unchanged — but those docs still enter `changed_source_ids` and get re-ingested.

With this PR they now hit delete-before-add → Memgraph node removed → `graph_synced` still `true` → Phase 4 never rebuilds → **permanent graph loss until content changes**. Title/URL/author/metadata-only connector updates are common; this is a regression vs pre-PR (no incremental delete on this path).

Please fix before merge (any one is fine):
1. In pipeline: if `incremental` and `skip_graph`, do not `_delete_graph_node` (or clear `graph_synced` after deleting so Phase 4 rebuilds).
2. In the upsert sidecar branch: also clear `graph_synced` if delete-before-add is now the connector contract.
3. In `connection_sync`: after ingesting sidecar/changed docs with `skip_graph`, force those source_ids back to `graph_synced=false`.

### Non-blocking
- Assert `incremental is True` in `test_sync_server_force_full` so the `not force_full` inversion can’t regress.
- Optionally append `sanitize_error(str(e))` into the persist-failure `errors_list` string (today only the generic message is stored).
- Worth a one-line ops note that `force_full` no longer rebuilds Qdrant for unchanged docs (intentional per #461).

cycterna2222288888-ai:
Thanks — the graph-loss analysis holds. Confirmed the chain in code:

`connection_sync` Phase 2 → `ingest_documents(incremental=True, skip_graph=True)` → per-doc `delete_by_doc_labels` returns >0 → `_delete_graph_node` (`pipeline.py:223`), while `upsert_raw_documents` leaves `graph_synced=true` on a sidecar-only refresh (#440) → `get_unsynced_documents(target="graph")` filters `NOT graph_synced` → Phase 4 never rebuilds the node.

### Fix — option 3 (f834a4f)

New `connection_sync` Phase 3b, right after the `skip_graph` ingest: clears `graph_synced` for the re-ingested `source_id`s via a new `PostgresStore.mark_documents_graph_unsynced_by_source`, so the Phase 4 sweeper rebuilds the deleted nodes. No-op for `new` / content-changed rows (`upsert_raw_documents` already left those `false`), so only the sidecar rows actually flip. Local to `connection_sync` — `ingest_documents` and its other callers are untouched.

### Regression test

`test_phase2_sidecar_refresh_reenqueues_graph_extraction` — feeds the sidecar-drift shape (`unchanged=1`, `changed_source_ids=["J-1"]`), asserts the re-ingest runs `skip_graph=True` **and** that `mark_documents_graph_unsynced_by_source(source_ids=["J-1"])` fires.

### Non-blocking — all done

- `test_sync_server_force_full` now asserts `incremental is True`.
- Persist failure: `sanitize_error(str(e))` is captured into `errors_list` → the `sync_logs` row; test 3 asserts it.
- PR description: added the note that a connector `force_full` no longer rewrites Qdrant for unchanged docs (intentional per #461).

CI green.

cycterna2222288888-ai:
Quick note — the fix for the graph-node-loss regression landed already (commit from 2026-09-08 18:41 UTC, before my analysis comment above), and CI's been green since. Your CHANGES_REQUESTED review predates that commit. Could you re-review when you get a chance?

PR #483, opened 2026-09-09
memory: validate importance_score and push tag filter into SQL

### LINKED ISSUE #455: memory_store accepts out-of-range importance_score; memory_list tag filter applied per-page with tag-unfiltered total
### What happened?

Two contract defects in the agent‑memory MCP tools.

---

**1. `metronix_memory_store` does not validate the `importance_score` range.**

The tool description and `docs/MCP_API.md` specify `importance_score` as `0.0..1.0` (default `0.5`). `metronix_memory_store` (`src/metronix/mcp/tools/memory_store.py`) does:

```python
importance_score=float(importance_score),
```

with no bounds check. `MemoryRecord` (`src/metronix/core/models.py`) is a plain dataclass — `importance_score: float = 0.5`, no validator — and there is no CHECK constraint on the column. `MemoryService.save` persisted a record with `importance_score = 999.0` and returned it unchanged.

`importance_score` is read back as `graph_score` in `MemorySearchService.hybrid_search` (`merged[record_id].graph_score = importance`), so an out‑of‑range value feeds ranking directly: `score += weights.graph * graph_score`, i.e. `0.3 * 999` for that record on every query.

Non‑numeric input (`float("abc")`) raises `ValueError`, which the tool's outer handler returns as `INTERNAL_ERROR` rather than `INVALID_PARAMS`.

---

**2. `metronix_memory_list` applies the `tags` filter per‑page, and `total` ignores it.**

`metronix_memory_list` (`src/metronix/mcp/tools/memory_list.py`):

```python
records = await service.pg_store.list_records(ws_id, ..., limit=limit, offset=offset)   # no tag filter in SQL
total   = await service.pg_store.count_records(ws_id, ...)                               # no tag filter
...
if tags:
    tag_set = set(tags)
    records = [r for r in records if tag_set & set(r.tags)]                              # post-filter, this page only
```

`list_records` / `count_records` accept `scope`, `kind_filter`, `status` but not `tags`. The tag filter runs in Python over the already‑paginated page, and `total` is the tag‑unfiltered count.

On agent `peter-work` (15 records), tag `decision` is on 5 records:

- `list_records(limit=3, offset=0)` → 3 records; after the `decision` post‑filter → 1.
- `count_records()` → 15.

So `metronix_memory_list(agent_id="peter-work", tags=["decision"], limit=3)` returns `{count: 1, total: 15}`. A client paginating a tag‑filtered listing gets a `total` that does not reflect the filter (15, not 5) and pages that are mostly empty, with matching records spread across offsets `0, 3, 6, 9, 12`. A client that stops when a page returns `count == 0`, or sizes its loop from `total`, does not retrieve all tagged records.

`docs/MCP_API.md` states: "`total` is the `count_records(status=...)` value scoped to the same filter — pagination stays honest."

### Expected

1. `metronix_memory_store`: reject `importance_score` outside `0.0..1.0` (and non‑numeric input) with `INVALID_PARAMS`, or document and apply clamping.
2. `metronix_memory_list`: push `tags` into `list_records` / `count_records` so pagination and `total` are consistent with the filter; or, if the post‑filter stays, document that `tags` is incompatible with pagination and that `total` excludes it.

### Steps to reproduce

`importance_score` (service layer; delete the probe record afterwards):

```python
import asyncio
from metronix.core.models import MemoryRecord, MemoryScope, MemoryKind
from metronix.mcp.tools._memory_deps import build_memory_service_for_workspace

async def go():
    svc = await build_memory_service_for_workspace("MTRNIX")
    rec = MemoryRecord(workspace_id="MTRNIX", agent_id="probe", scope=MemoryScope.PER_AGENT,
                       kind=MemoryKind.FACT, content="range test", importance_score=999.0)
    saved = await svc.save("MTRNIX", rec)
    print("stored importance_score:", saved.importance_score)   # -> 999.0
    await svc.delete("MTRNIX", saved.id, agent_id="probe")

asyncio.run(go())
```

`memory_list` tags / pagination (through the MCP tool, given an authorized principal — #424):

1. Store ~15 records for one agent; put the same tag on 5 of them, spread across creation order.
2. `metronix_memory_list(agent_id="<agent>", workspace_id="MTRNIX", tags=["<tag>"], limit=3, offset=0)`.
3. `count` reflects only the tagged records within the first 3 rows (0–1); `total` is the full unfiltered count (15).
4. Page `offset = 3, 6, 9, 12`; the 5 tagged records appear one or two at a time; no page's `count` and no `total` indicates how many tagged records exist.

Equivalent at the store layer (no principal needed):

```python
recs = await svc.pg_store.list_records("MTRNIX", agent_id="<agent>", limit=3, offset=0)
print(len([r for r in recs if "<tag>" in r.tags]))                   # 0 or 1
print(await svc.pg_store.count_records("MTRNIX", agent_id="<agent>"))  # 15
```

### Relevant logs / output

```
# importance_score
stored importance_score: 999.0

# memory_list tags/pagination (store layer, agent=peter-work, tag=decision)
records with tag "decision" total ............. 5
list_records(limit=3, offset=0), tag matches .. 1
count_records() ............................... 15
```

### Version / commit

`main` @ `ca50e5a`, local `docker compose` stack.

### Environment

macOS 15.6, Docker 29.6.2 / Compose v2. Stack via `docker compose up -d --build` from the repo `docker-compose.yml`. `.env`: `AUTH_ENABLED=false`, embeddings `nomic-embed-text`, freshness disabled.

Moviw:
## Summary

Two contract defects in the agent-memory MCP tools, both from #455.

**1. `importance_score` was unbounded on the MCP surface.** The REST route already enforces the documented range with `importance_score: float = Field(0.5, ge=0.0, le=1.0)`, so the accurate framing is not "nothing validates this" but "the two MCP entry points drifted from a contract the rest of the codebase honours". A stored `999.0` does not fail loudly — `MemorySearchService.hybrid_search` reads it back as `graph_score` and multiplies it into the ranking sum, so it silently dominates results.

The issue names `memory_store`; `memory_batch_store` had the same bare `float(importance_score)`, so both are fixed. Validation lives in `_memory_utils.py` next to `scope_from_str` and `validate_kind`, raising `ValueError` for callers to map to `INVALID_PARAMS`, matching the existing idiom.

Non-numeric input is also fixed as a side effect: `float("abc")` previously raised inside the broad `except Exception` and `handle_tool_error` classified it `INTERNAL_ERROR`, so a client got a server error for its own bad argument. It is now `INVALID_PARAMS`.

**2. The `tags` filter broke pagination.** It ran in Python over the already-paginated page while `total` came from a tag-unfiltered `count_records`, so `metronix_memory_list(tags=["decision"], limit=3)` returned `{count: 1, total: 15}` with matches scattered across offsets. `docs/MCP_API.md` promises "pagination stays honest". `tags` is now a push-down filter on `list_records` and `count_records`.

## Related issue

Closes #455

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / internal change
- [ ] Documentation
- [ ] CI / tooling

## Design notes

**Why validation is at the ingress boundary and not in `MemoryRecord`.** A raising `__post_init__` would be the smaller diff and would cover every caller, but that dataclass is also the shape rows are read back into (`memory_postgres.py:112`, `memory_redis.py:55`). Records already written by this bug would become unreadable and undeletable — trading a silent ranking skew for a hard read failure. The two MCP tools are the untrusted ingress; that is where the guard belongs.

**Why `?|` and not the `jsonb_array_elements_text` idiom already in this file.** The existing form at `memory_postgres.py:713` raises `cannot extract elements from a scalar` on any row whose `tags` is a JSON scalar rather than an array — verified against a real server, see below. `?|` is safe on `[]`, JSON `null` and SQL `NULL`. `@>` was rejected outright: it means "contains every tag", not "any overlap". The explicit `CAST(... AS text[])` follows the note at `memory_postgres.py:1282` about asyncpg array element inference.

## Validation

Run against a real PostgreSQL with all 34 Alembic migrations applied.

```
alembic upgrade head                                    # 34 migrations, exit 0
pytest tests/unit/mcp tests/unit/storage -m "not integration"
                                                        # 208 passed (baseline 202)
pytest tests/unit/mcp/test_memory_utils.py              # 23 passed (18 new)
pytest tests/unit/storage/test_memory_postgres_tags_filter.py
                                                        # 6 passed (new file)
ruff check src/ tests/                                  # All checks passed
ruff format --check src/ tests/                         # 724 files already formatted
mypy <the 5 touched files>                              # 0 errors each
```

`mypy src/metronix/` reports 3283 pre-existing errors across 111 files; the five files this PR touches are clean individually.

The `?|` semantics were checked directly, through SQLAlchemy `text()` and asyncpg rather than by inspection, because a literal `?` in the SQL was the main risk:

| `tag_list` | rows returned |
| --- | --- |
| `["decision"]` | 1, 3 |
| `["decision", "work"]` | 1, 2, 3 |
| `["nope"]` | none |
| `[]` | none, no error |

Case-sensitive, matching the Python `set` intersection it replaces; safe on `[]`, JSON `null` and SQL `NULL` rows; `count_records` agrees with `list_records` on the same filter.

**Mutation-checked** — each change reverted individually to confirm the tests catch it, then restored:

| Mutation | Result |
| --- | --- |
| Drop the range check | 8 failed |
| Catch only `ValueError`, not `TypeError` | 3 failed |
| Swap `?\|` for `@>` | 3 failed |
| Add the clause to `list_records` but not `count_records` | 1 failed |

## Not covered

My local server is PostgreSQL 18.6; CI runs `postgres:16-alpine`. `?|` behaviour is unchanged between those versions, but the numbers above come from 18.6.

Tests needing Neo4j, Qdrant or Redis were not run locally — only the PostgreSQL-backed subset.

### DIFF
--- a/src/metronix/mcp/tools/_memory_utils.py
+++ b/src/metronix/mcp/tools/_memory_utils.py
@@ -54,6 +54,29 @@ def parse_status_filter(
     return out
 
 
+def validate_importance_score(value: float) -> float:
+    """Coerce ``importance_score`` to a float inside the documented 0.0..1.0 range.
+
+    Raises ``ValueError`` when the value is not numeric, is not finite, or falls
+    outside the range. ``importance_score`` is read back as ``graph_score`` in
+    ``MemorySearchService.hybrid_search`` and multiplied into the ranking sum, so
+    an out-of-range score does not fail loudly — it silently dominates ranking.
+    The REST surface already enforces this bound through its Pydantic field
+    (``Field(0.5, ge=0.0, le=1.0)``); this keeps the MCP tools on the same
+    contract.
+    """
+    try:
+        score = float(value)
+    except (TypeError, ValueError) as exc:
+        msg = f"importance_score must be a number, got {value!r}"
+        raise ValueError(msg) from exc
+    # Rejects NaN as a side effect: every comparison against NaN is False.
+    if not 0.0 <= score <= 1.0:
+        msg = f"importance_score must be between 0.0 and 1.0, got {score}"
+        raise ValueError(msg)
+    return score
+
+
 def validate_kind(kind: str | None) -> MemoryKind | None:
     """Convert an optional kind string to ``MemoryKind``.
 
--- a/src/metronix/mcp/tools/memory_batch_store.py
+++ b/src/metronix/mcp/tools/memory_batch_store.py
@@ -11,7 +11,7 @@ from metronix.core.utils import is_valid_agent_id
 from metronix.mcp.errors import ErrorCode, MCPError, handle_tool_error
 from metronix.mcp.server import mcp
 from metronix.mcp.tools import _memory_deps
-from metronix.mcp.tools._memory_utils import scope_from_str
+from metronix.mcp.tools._memory_utils import scope_from_str, validate_importance_score
 from metronix.mcp.tools.models import MemoryBatchStoreResponse, MemoryBatchStoreResult
 
 logger = structlog.get_logger(__name__)
@@ -100,6 +100,16 @@ async def metronix_memory_batch_store(
                 ).to_dict(),
             }
 
+        try:
+            validated_importance = validate_importance_score(importance_score)
+        except ValueError as exc:
+            return {
+                "error": MCPError(
+                    code=ErrorCode.INVALID_PARAMS,
+                    message=f"metronix_memory_batch_store: {exc}",
+                ).to_dict(),
+            }
+
         from metronix.mcp.config import resolve_workspace_id
         from metronix.mcp.tools._agent_access import require_agent_access
 
@@ -130,7 +140,7 @@ async def metronix_memory_batch_store(
                     source_type=source_type,
                     content=str(content),
                     tags=list(tags),
-                    importance_score=float(importance_score),
+                    importance_score=validated_importance,
                     session_id=session_id,
                 )
                 new_id = record.id
--- a/src/metronix/mcp/tools/memory_list.py
+++ b/src/metronix/mcp/tools/memory_list.py
@@ -113,6 +113,7 @@ async def metronix_memory_list(
             scope=scope_enum,
             kind_filter=kind_filter,
             status=status_filter,
+            tags=tags,
             limit=limit,
             offset=offset,
         )
@@ -122,12 +123,8 @@ async def metronix_memory_list(
             scope=scope_enum,
             kind_filter=kind_filter,
             status=status_filter,
+            tags=tags,
         )
-
-        # Post-filter by tags (intersection — record must have at least one matching tag)
-        if tags:
-            tag_set = set(tags)
-            records = [r for r in records if tag_set & set(r.tags)]
 
         dto_list = [
             MemoryRecordDTO(
--- a/src/metronix/mcp/tools/memory_store.py
+++ b/src/metronix/mcp/tools/memory_store.py
@@ -11,7 +11,11 @@ from metronix.core.utils import is_valid_agent_id
 from metronix.mcp.errors import ErrorCode, MCPError, handle_tool_error
 from metronix.mcp.server import mcp
 from metronix.mcp.tools import _memory_deps
-from metronix.mcp.tools._memory_utils import scope_from_str, validate_kind
+from metronix.mcp.tools._memory_utils import (
+    scope_from_str,
+    validate_importance_score,
+    validate_kind,
+)
 from metronix.mcp.tools.models import MemoryStoreResponse
 
 logger = structlog.get_logger(__name__)
@@ -88,6 +92,16 @@ async def metronix_memory_store(
                 ).to_dict(),
             }
 
+        try:
+            validated_importance = validate_importance_score(importance_score)
+        except ValueError as exc:
+            return {
+                "error": MCPError(
+                    code=ErrorCode.INVALID_PARAMS,
+                    message=f"metronix_memory_store: {exc}",
+                ).to_dict(),
+            }
+
         if scope_enum == MemoryScope.SESSION and not session_id:
             return {
                 "error": MCPError(
@@ -109,7 +123,7 @@ async def metronix_memory_store(
             source_type=source_type,
             content=content,
             tags=list(tags) if tags else [],
-            importance_score=float(importance_score),
+            importance_score=validated_importance,
             session_id=session_id,
         )
         new_id = record.id
--- a/src/metronix/storage/memory_postgres.py
+++ b/src/metronix/storage/memory_postgres.py
@@ -283,6 +283,7 @@ class MemoryPostgresStore:
         kind_filter: list[MemoryKind] | None = None,
         source_type_filter: list[str] | None = None,
         status: list[LifecycleStatus] | None = None,
+        tags: list[str] | None = None,
         lifetime: str = "all",
         limit: int = 100,
         offset: int = 0,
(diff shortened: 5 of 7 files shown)


cycterna2222288888-ai (changes requested):
Storage-layer fix looks correct — pushing the tag filter into `list_records`/`count_records` via `?|` fixes the pagination/total mismatch from #455, and `test_memory_postgres_tags_filter.py` covers the SQL clause well.

The CI failure is real, not flaky: `tests/unit/test_memory_list.py::TestMemoryList::test_tags_post_filter` (pre-existing, untouched by this diff) asserts the *old* bug as the expected contract — `assert out["total"] == 2` with the comment `# total reflects DB count (before tag filter)`. Its `_mock_service` doesn't implement tag filtering at all; it just returns both fixture records regardless of `tags=`. That worked before because `memory_list.py` filtered the page in Python after the mock returned. Now that the Python-side filter is removed, the mock returns both records unfiltered and `count` is 2, not the asserted 1.

This test needs to go or be rewritten — it's currently pinning the exact behavior this PR closes. A replacement worth having: assert that `metronix_memory_list` actually forwards `tags` into the service/store call, since the new tests only cover the SQL-clause level in `MemoryPostgresStore` — nothing currently checks the MCP tool layer passes `tags` through rather than dropping it.

Moviw:
Good catch, thanks — you're right, `test_tags_post_filter` was still pinning the pre-fix behavior (the mock returned both fixture records unfiltered and asserted `total == 2`), which only worked because `memory_list.py` used to filter the page in Python after the mock returned. Since that Python-side filter is gone now that tags are pushed into the SQL clause, the mock's lack of filtering was masking the real bug this PR fixes.

Pushed 325f0cf: rewrote it as `test_tags_forwarded_to_store`, which now asserts `metronix_memory_list` actually forwards `tags=["important"]` into both `list_records` and `count_records` (checked via `call_args`), rather than re-asserting the old unfiltered-mock total. This covers exactly the gap you flagged — the MCP tool layer passing `tags` through instead of dropping it — while `test_memory_postgres_tags_filter.py` continues to cover the SQL-clause level.

`tests/unit/test_memory_list.py` (4/4) and `tests/unit/storage/test_memory_postgres_tags_filter.py` pass locally against the new commit.

Moviw:
Good catch, thanks — pushed 325f0cf.

`test_tags_post_filter` is gone; replaced with `test_tags_forwarded_to_store`, which does what you suggested: asserts `metronix_memory_list` actually forwards `tags=["important"]` into both `list_records` and `count_records` (`call_args` on the mocked `pg_store` methods), rather than asserting on the old Python-side-filter behavior. The mock now stands in for a store that already applied the SQL-level filter (one matching record, `total=1`), consistent with what `MemoryPostgresStore` actually does post-fix.

Ran the full unit suite locally (`tests/unit`, excluding a few pre-existing env-only failures unrelated to this diff — missing `discord`/`slack` extras and no local Postgres for `test_connector_state.py`): 1092 passed, 5 skipped, 0 new failures.

cycterna2222288888-ai (changes requested):
Основная часть (SQL push-down тегов и `validate_importance_score`) сделана аккуратно, детали проверил — оператор, параметризация, guard, тесты покрывают нужные случаи.

Но валидация не подключена в `memory_update.py` (строки 104 и 128) — `importance_score` там идёт без проверки, хотя это тот же путь, что описан как мотивация PR: значение читается как `graph_score` и молча доминирует в ранжировании. Через `metronix_memory_update` можно протащить `-5` или `999` в существующую запись, обходя всю проделанную работу.

Раз PR заявляет цель "validate importance_score" целиком — это незакрытая часть заявленного скоупа, а не отдельный тикет. Прошу добавить проверку и туда.

Moviw:
Good catch, thanks — pushed 824154d.

`metronix_memory_update` now validates `importance_score` through the same `validate_importance_score()` used by `metronix_memory_store`/`memory_batch_store`, before it reaches `pg_store.update()` and before it reaches the Qdrant payload update (both spots you flagged). An out-of-range value on either path now returns `INVALID_PARAMS` instead of getting written.

Added `test_out_of_range_importance_score_rejected` in `tests/unit/test_memory_update.py`, asserting the update is rejected before the service is even built (`build_memory_service_for_workspace` never called). Ran the full local suite again: `ruff check` and `mypy` clean on the changed file, 27 passed on the memory_update/memory_store/memory_batch_store subset, no regressions.

cycterna2222288888-ai (approved):
Оба места закрыты: `memory_update.py` теперь валидирует `importance_score` через `validate_importance_score()` до `pg_store.update()` и до записи в Qdrant payload, с явным тестом на отказ до обращения к сервису (`build_memory_service_for_workspace` не вызывается на невалидном значении). Оба замечания из ревью закрыты аккуратно.

PR #485, opened 2026-09-13
benchmarks: parse env files with Compose inline-comment semantics

### LINKED ISSUE #470: Benchmark env loader (_parse_env_file) strips no inline comments and reads the repo-root .env — diverges from the Compose env_file parser

godedok:
Fixes #470.

## What

Both benchmark env loaders (`benchmarks/locomo/scripts/env_config.py`, `benchmarks/longmemeval/scripts/env_config.py`) parsed env files with a hand-rolled parser that had no `#` handling, so `KEY=value # note` kept the comment in the value and `KEY= # note` became `# note` — diverging from the Compose `env_file` parser applied to the same repo-root `.env`.

`_parse_env_file` now normalizes each value through a `_clean_env_value` helper that follows Compose semantics:

- an unquoted value is cut at the first ` #`; a `#` not preceded by a space stays literal (URL fragments, `value#tag`, …)
- a value that is only a comment resolves to an empty string
- quoted values keep `#` inside the quotes, drop one pair of matching outer quotes, and discard a trailing comment after the closing quote

The repo-root `.env` read stays as is (it intentionally fills install credentials) and is now documented in both `load_dotenv` docstrings — those values are parsed with the same Compose semantics.

## Testing

- New `tests/unit/benchmarks/locomo/test_locomo_env_config.py` (the locomo loader had no parser coverage) and extended the longmemeval suite with the exact repro cases from the issue plus quoted / hash-without-space cases.
- All 8 env-config tests pass locally; ruff check + format clean.


toomij99 (approved):
LGTM — matches #470. Compose-style ` #` comment stripping + quoted hashes kept, plus good table-driven tests on both LoCoMo and LongMemEval loaders. CI green.

Nit (non-blocking): `_clean_env_value` is duplicated in both env_config.py files. Fine for now; extract only if a third copy appears.

`C=#immediate` → empty is consistent with the stated Compose rule (value starts with `#`).

Issue #489, opened 2026-09-14
compare.py never computes error_count for LongMemEval — comparator silently reports 0 errors on a degraded leg


cycterna2222288888-ai:
## `compare.py` never computes `error_count` for LongMemEval — comparator silently reports 0 errors on a degraded leg

`benchmarks/_shared/compare.py::metrics()` computes `error_count` correctly for LoCoMo but not for LongMemEval:

```python
error_count = 0                                    # line 105

if run.benchmark == "locomo":
    ev = run.answer_eval or {}
    ...
    error_count = int(ev.get("error_count", 0) or 0)   # line 113 — set
else:  # longmemeval                                    # line 114
    recall_report = run.retrieval_eval
    latency_report = run.retrieval_eval.get("latency") if run.retrieval_eval else None
    answer_metric, answer_metric_name = _longmemeval_answer_metric(run)
    # error_count is never touched here — stays 0
```

`error_count` is later used in `check_comparable()`:

```python
if run_metrics["error_count"]:
    reasons.append(f"{label} run has benchmark errors — treat it as a failed leg")
```

For LongMemEval this check is dead code — it can never fire, no matter how many questions actually failed.

### How this was found

Running the full LongMemEval-S set (500 questions) on `feat/benchmark-reproducible-harness @ 63cad02`, 79/500 rows (15.8%) have `hypothesis` starting with `"Error: unhandled errors in a TaskGroup (1 sub-exception)"` (root cause: `metronix_memory_store: context length exceeded` on ingest — see a possible separate issue for that). These rows carry no `recall_gate_eligible` field, so `aggregate_recall` correctly excludes them from the recall computation. But the LLM judge grades every row's `hypothesis` against the reference answer regardless — all 79 were graded `false` (an "Error: ..." string obviously never matches the reference), which silently deflates the reported accuracy: 366/500 = 0.732 reported vs. 366/421 = 0.869 on the 421 rows that were genuinely attempted.

`compare_runs.py results/off.jsonl results/on.jsonl` on this pair reports `error_count: 0` for both legs and would happily call them comparable — exactly the case the LoCoMo branch's identical check is designed to catch.

### Fix

[...]

PR #491, opened 2026-09-15
fix(longmemeval): guard chat_complete against None completion (#490)

### LINKED ISSUE #490: chat_complete doesn't guard completion is None before .choices[0] — unretryable TypeError, multi-minute stalls, lost answers
## `chat_complete` in `run_benchmark.py` doesn't guard against `completion is None`; the resulting `TypeError` isn't retryable — causes multi-minute stalls and lost answers

`benchmarks/longmemeval/scripts/run_benchmark.py`:

```python
@backoff.on_exception(backoff.expo, (openai.RateLimitError, openai.APIError), max_tries=8)  # line 163
def chat_complete(client: OpenAI, *, model: str, user_message: str) -> str:
    completion = client.chat.completions.create(...)
    return completion.choices[0].message.content.strip()   # line 174
```

When the provider (DeepSeek here) returns a response the `openai` client parses as `completion = None` (observed right after an HTTP 200 in the log — not a rate limit or API error), line 174 raises `TypeError: 'NoneType' object is not subscriptable`. That exception type is not in `backoff`'s retry tuple (`openai.RateLimitError, openai.APIError`), so it is never retried at the `chat_complete` level.

It propagates up through `process_question` to the per-question `try/except Exception` in `run_benchmark()`'s loop, which catches it, logs it, and records `hypothesis = f"Error: {exc}"` for that question — a bare error string with no retrieval data, which the LLM judge will always grade as wrong (see #489 for the accuracy-contamination consequence of that pattern).

### Cost observed

On a full LongMemEval-S flag-on run (`feat/benchmark-reproducible-harness @ 63cad02`), two consecutive questions each took 15-18 minutes of wall-clock time before this exception was hit (`993da5e2`: 1,092,975 ms; `a3045048`: 1,036,072 ms) — both with `ingest_ms: null`, consistent with the full per-question pipeline (ingest + search) re-running before the answer step failed. This is what produced a temporary tqdm ETA spike to 27+ hours during that stretch. 5 occurrences seen so far in one run.

### Suggested fix

- Raise on `completion is None` (or `not completion.choices`) with an exception type that's actually useful to the caller — either add it to `chat_complete`'s `backoff.on_exception` tuple so it gets retried in place, or raise something distinguishable from a generic `TypeError` so the per-question handler can log it more usefully than a bare `'NoneType' object is not subscriptable`.
- Worth checking whether this is a `provider quirk` (DeepSeek returning a malformed 200) the `openai` SDK doesn't surface as an `APIError` — if so it may need its own guard rather than relying on the SDK's exception hierarchy.

### Version

`feat/benchmark-reproducible-harness @ 63cad02`

toomij99:
## Summary
Closes #490.

`chat_complete` in the LongMemEval runner crashed with `TypeError: 'NoneType' object is not subscriptable` when a provider (observed with DeepSeek after HTTP 200) returned a completion the OpenAI client parsed as `None`. That exception is not in the backoff retry tuple, so the failure bubbled to the per-question handler as a bare error string after the full ingest+search pipeline had already run (multi-minute stalls).

## Root cause
```python
completion = client.chat.completions.create(...)
return completion.choices[0].message.content.strip()  # crashes when completion is None
```

## Fix
Mirror the LoCoMo runner pattern already on main:
- Add `EmptyChatCompletionError`
- Guard `completion is None`, empty `choices`, and empty/non-string content
- Include `EmptyChatCompletionError` in `@backoff.on_exception` so these cases retry in place (up to 8 tries)

## Tests
- `test_chat_complete_retries_when_completion_is_none` — None then success
- `test_chat_complete_retries_empty_choices` — empty choices then success
- `test_chat_complete_reports_none_completion_after_retries` — exhausted retries raise `EmptyChatCompletionError`

Locally: `ruff format --line-length 99` + `ruff check` clean on touched files; the three new tests passed.

## Scope
LongMemEval only — does not touch #489/#486/#488.


toomij99:
@cycterna2222288888-ai quick ask: branch protection needs your Approve (`toomij99` can’t self-approve). Diff is #490 — treat None/empty chat completions as retryable in LongMemEval `chat_complete`. Internal review LGTM; squash-merge when CI green.

https://github.com/mtrnix/metronix-memory/pull/491

toomij99:
**Re-check (toomij99 — cannot self-approve)**

LGTM on HEAD `bc9b2fd`. `message is None` guard is in place before `.content`; still raises retryable `EmptyChatCompletionError`. Ruff green; pytest still running. Squash-merge after Cycterna Approve + green CI.

cycterna2222288888-ai (approved):
Фикс проверен, закрывает #490: `completion`/`choices`/`message`/`content` последовательно гардятся, новый `EmptyChatCompletionError` добавлен в retryable-кортеж backoff вместо необработанного `TypeError`. Тесты покрывают все 4 сценария (None-completion, пустые choices, None-message, исчерпание ретраев).

PR #492, opened 2026-09-15
fix(benchmarks): count LongMemEval Error: hypotheses in compare.error_count

### LINKED ISSUE #489: compare.py never computes error_count for LongMemEval — comparator silently reports 0 errors on a degraded leg

toomij99:
Closes #489.

## Problem
`benchmarks/_shared/compare.py::metrics()` set `error_count` for LoCoMo from `.eval.json`, but left it at `0` for LongMemEval. `check_comparable()` therefore never treated LongMemEval infra failures as a failed leg — even when ~15–18% of `answers.jsonl` rows were `Error: …` hypotheses.

## Fix
- Add `_longmemeval_error_count()` that counts JSONL rows whose `hypothesis` starts with `"Error:"` (same rule as `benchmarks/locomo/scripts/evaluate.py`).
- Assign that count in the LongMemEval branch of `metrics()`.

## Tests
- `test_metrics_longmemeval_counts_error_hypotheses` — asserts `error_count == 2` for mixed ok/`Error:` rows (not stuck at 0).
- `test_longmemeval_errors_block_comparison` — asserts a leg with `Error:` hypotheses is not comparable.

## Notes
- Branched from `feat/benchmark-reproducible-harness` (not `main`): `benchmarks/_shared/compare.py` exists only on the harness branch (same base as #486).
- Scope limited to #489; did not touch #486/#488/#491.
- `ruff format --line-length 99` + `ruff check` clean on touched files (project `pyproject.toml` config).
- Not merged.


cycterna2222288888-ai:
Heads up — #489 is already closed on this same base branch: commit `244ba58` on `feat/benchmark-reproducible-harness`, landed as part of #493. Same fix, essentially line-for-line: `_longmemeval_error_count()` in `compare.py`, same `hypothesis.startswith("Error:")` rule, same insertion point in `metrics()`. Test coverage is equivalent too (`test_metrics_longmemeval_counts_error_hypotheses` + a comparability-blocking test on both sides, just named differently). Neither implementation is meaningfully better than the other on this part — genuinely parallel, independent fixes for the same bug.

Since `feat/benchmark-reproducible-harness` already has the fix, merging this as-is will hit a real conflict (or a silent no-op merge) on `compare.py` / `test_compare.py` — recommend closing this as superseded by #493/244ba58 rather than resolving that conflict for a diff that's already landed.

**Separately, and this matters regardless of the #489 overlap:** the diff here also touches `scripts/run_eval.py`, and that part looks like it's carrying an unintended regression, not a #489-related change. It strips ~300 lines — the actual `run_eval`/`save_results`/`compare_results`/`show_history` implementations — and replaces `main()`'s body with:

```python
raise SystemExit(
    "Full eval run body not restored on this branch tip yet; "
    "--help works. Restore CLI body from 98f021805ca36d3c6ef4308023bd43eb0c16b372 "
    "keeping the ImportError-based benchmark_qed stub."
)
```

That reads like a local branch state that had `run_eval.py` mid-restore from a bad rebase/cherry-pick, checked in as-is. If this merges, `python scripts/run_eval.py` (and `make eval`/`make eval-save`/`make eval-compare`/`make eval-history`) stop working entirely — that's unrelated to #489 and worth fixing (or dropping from this PR) independent of what happens with the duplicate fix above.

The `.github/workflows/tests.yml` change (tee-ing pytest output and annotating `FAILED`/`ERROR` lines as `::error::`) looks like a genuinely nice, unrelated CI improvement — worth keeping/extracting into its own PR if this one closes.

PR #503, opened 2026-09-21
fix(benchmarks): name pre-harness runs missing query_set (#484)

### LINKED ISSUE #484: benchmarks: compare_runs.py cannot load a run - query_set.json sidecar never written

toomij99:
Closes #484 (UX only — root cause was pre-harness artifacts, not a harness bug). Supersedes #486.

## What
When `load_run` finds a `.manifest.json` but no `.query_set.json`, raise a specific `FileNotFoundError` explaining the run predates the query-set sidecar and must be re-run on the reproducible harness — instead of the generic "missing both sidecars" message.

## Why
Pre-harness LoCoMo runs wrote only the manifest. Those artifacts look healthy until compare fails with a confusing error. This is a UX clarification only; no harness or generator changes.

## Tests
- `test_load_run_manifest_without_query_set_reports_pre_harness_run`
- `test_load_run_query_set_without_manifest_reports_both`


cycterna2222288888-ai (approved):
Checked the full diff (2 files, +29/-0, no side changes): the new branch in `load_run()` only fires when the manifest is present and the query set is absent, the original 'missing both sidecars' path is untouched, and the message matches the scenario.

Tests: with the fix, test_compare.py is 34 passed. With compare.py reverted to main, `test_load_run_manifest_without_query_set_reports_pre_harness_run` fails on the expected regex, so it really exercises the fix. `..._query_set_without_manifest_reports_both` passes on both, i.e. it pins the unchanged fallback, which is what it claims.

I also hit this case for real: my pre-harness LoCoMo PPR runs (`ppr-off-full`, `ppr-on-full`) have a .manifest.json but no .query_set.json. Through this branch `load_run` now raises the new 'predates the query-set sidecar ... re-run' error for both.

Non-blocking nit: `_load_json` also returns None for a query_set file that exists but isn't a JSON object, in which case the message would say 'no .query_set.json'. Edge case, fine to leave.

